PR #22, opened 2021-07-28
initialize fetch trait

Sczlog:
1. first version of fetch trait.
2. example of lazy and non-lazy.

known-issue:
cannot detect internal change of headers with expression


Sczlog:
I wrote an example for listening header with expression change, modified `store.ts`  at this branch https://github.com/Sczlog/meta-ui/tree/expressionobject

Yuyz0112:
Excellent work!

PR #33, opened 2021-08-10
impl expression evalutaion v2

Yuyz0112:
Also, migrate components and traits.


packages/runtime/src/components/chakra-ui/NumberInput.tsx
```diff
@@ -55,10 +55,11 @@ const NumberInput: ComponentImplementation<{
   size,
   customerIncrement,
   customerDecrement,
-  mergeState
+  mergeState,
```
Sczlog, line 58: most change origin from formating, should we add a proper eslintrc/prettierc file here?

Yuyz0112: https://github.com/webzard-io/meta-ui/issues/34

PR #39, opened 2021-08-12
implement router v1

Sczlog:
### Basic:
use wouter as basic router component to provide functional router. https://github.com/molefrog/wouter
use PathToRegexp as path resolver. https://github.com/pillarjs/path-to-regexp
support both history api and hash as router.

### Implementation:
1. implement router component:
The function of the component is more like a `Switch Component`
We need to defined the policy for the router inside its property.
Router will match the policy one by one until the first one is matched, and render the component  by its id.
The component must has a route trait which routerId point the router component.
2. implement route trait:
The trait will tell renderer not to render it as a top level component, like slot trait but to parse it into routers component registered for another component, by the given router id. The component will be rendered only if its router is rendered and the path satisfied the policy.

### TODO:
- [x] 1. Add default and error component as fallback component.
- [x] 2. Add `exact`, `strict` for route policy or route trait.
- [x] ~3. Add support for nested-router with relative path. (We can use nested-router now, but for the nested-router, they cannot use relative path)~
- [x] 4. Parse parameter for route component


packages/runtime/example/router/index.html
```diff
@@ -0,0 +1,159 @@
...
+          components: [
+            {
+              id: "router",
+              type: "core/v1/router",
+              properties: {
+                routerPolicy: [
+                  {
+                    type: "route",
+                    path: "/react",
+                    cid: "text_react",
```
Yuyz0112, line 28: How about change `cid` to slot name and use `slot` trait instead of `route` trait?

With this change, we can simplify what each component/trait does:
- the router component toggle slots by matching URL
- the slot trait helps layout arbitrary component

Sczlog: @Yuyz0112 
I just updated with slot trait.
but I feel a little bit werid about using slot in this situation.
And the second point means we can put multiple components into one matched slot?

tanbowensg:
LGTM

Yuyz0112 (approved):

Issue #42, opened 2021-08-24
new trait implementation


tanbowensg:
Chinese Version: https://thoughts.teambition.com/share/6124b6fdf1962900423f2e5e

- [x] New Trait Implementation
- [x] Refactor Validation Trait
- [x] Refactor Event Trait
- [x] Refactor State Trait
- [x] Refactor Hidden Trait
- [x] Global Event Handler
- [ ] Adaptor between Component & Trait

# Why we need new implementation
1. Trait is now implemented based on executing hooks in a loop, which is not the correct usage of hooks.
2. Dynamic creating component in Trait hook will cause unnecessary mount & unmount.  see demo on branch `validation-bug`

# New implementation

1. Components will received props from Traits. Components should implement some defined capabilities of components (such as how to deal with style, data, etc.)

2. Trait is an ordinary function, not a Hook.

3. Trait accepts the parameters from the Schema, and then returns an object. This object is implemented in accordance with defined capabilities of component.

4. Merge all the objects returned by Trait and pass them to the component.

<img width="657" alt="截屏2021-08-27 上午11 25 32" src="https://user-images.githubusercontent.com/12260952/131067154-30a705c8-8e55-48c4-8215-c2fc73d398c5.png">
# Trait restrictions

1. Each component must implement some interface to interact with traits

2. Hook cannot be used in Trait (or the number of Trait cannot be changed dynamically?)
3. Cannot use HOC in Trait

# How to implement the existing Trait?

1. Hidden

     1. Return style, use display: none

1. State

     1. Return data, and the component calls mergeState to bind it to the store

1. event

     1. Return the callback function. The timing of unsubscribe is controlled by the component.

tanbowensg:
> @tanbowensg So in the new design, a trait should only interact with components by using plain JS objects.
> 
> If some general function needs to add a UI/React element to the component, then it should be a component instead of a trait?

It can be a trait. For example, in form validation, we need a text to show error message. But the validation trait does not add a text element to input. It just merge error message to store. We need to add a text component explicitly in schema to show error message.
I'm not sure this is the best pratise so far because it will cause many template code.

Yuyz0112:
@tanbowensg agree

PR #47, opened 2021-08-30
Refactor trait new implementation

tanbowensg:
#42

### DIFF
--- a/packages/runtime/src/registry.tsx
+++ b/packages/runtime/src/registry.tsx
@@ -50,12 +50,13 @@ export type ComponentImplementation<T = any> = React.FC<
     subscribeMethods: SubscribeMethods;
     slotsMap: SlotsMap | undefined;
     style?: CSSProperties;
-    data?: unknown;
+    data?: Record<string, any>;
   }
 >;
 
 export type TraitImplementation<T = any> = (
   props: T & {
+    componentId: string;
     mergeState: MergeState;
     subscribeMethods: SubscribeMethods;
   }
--- a/packages/runtime/src/registry.tsx
+++ b/packages/runtime/src/registry.tsx
@@ -60,7 +60,6 @@ export type TraitImplementation<T = any> = (
   }
 ) => {
   props: any;
-  component?: React.FC;
 };
 
 class Registry {
--- a/packages/runtime/src/registry.tsx
+++ b/packages/runtime/src/registry.tsx
@@ -54,15 +54,20 @@ export type ComponentImplementation<T = any> = React.FC<
   }
 >;
 
+export type TraitResult = {
+  props: {
+    data?: unknown;
+    style?: CSSProperties;
+  } | null;
+};
+
 export type TraitImplementation<T = any> = (
   props: T & {
     componentId: string;
     mergeState: MergeState;
     subscribeMethods: SubscribeMethods;
   }
-) => {
-  props: any;
-};
+) => TraitResult;
 
 class Registry {
   components: Map<string, Map<string, ImplementedRuntimeComponent>> = new Map();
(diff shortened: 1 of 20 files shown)


Yuyz0112:
@tanbowensg Could you rebase the latest master then we can start review without looking into format diffs:)

tanbowensg:
> @tanbowensg Could you rebase the latest master then we can start review without looking into format diffs:)

Done.

packages/runtime/src/registry.tsx
```diff
@@ -49,18 +50,25 @@ export type ComponentImplementation<T = any> = React.FC<
...
     slotsMap: SlotsMap | undefined;
+    style?: CSSProperties;
+    data?: Record<string, any>;
   }
 >;
 
+export type TraitResult = {
+  props: {
+    data?: unknown;
+    style?: CSSProperties;
```
Yuyz0112, line 61: Looks good to me in this patch.

In the future, maybe we can have multiple styles in the trait results? With this, we can let components declare multiple customizable style slots.

packages/runtime/src/traits/core/fetch.tsx
```diff
@@ -1,8 +1,9 @@
-import { useCallback, useEffect, useMemo } from 'react';
 import { createTrait } from '@meta-ui/core';
 import { Static, Type } from '@sinclair/typebox';
 import { TraitImplementation } from '../../registry';
 
+let hasFetched = false;
```
Yuyz0112, line 5: Using a module-level variable will cause all the fetch traits to share this state?

packages/runtime/src/traits/core/state.tsx
```diff
@@ -1,30 +1,44 @@
...
+}> = ({ key, initialValue, componentId, mergeState, subscribeMethods }) => {
+  const hashId = `#${componentId}@${key}`;
+  let hasInitialized = HasInitializedMap.get(hashId);
+
+  if (!hasInitialized) {
     mergeState({ [key]: initialValue });
 
-    subscribeMethods({
-      setValue(value) {
+    const upperCaseKey = capitalizeFirstLetter(key);
```
Yuyz0112, line 20: not used in the latest implementation?

packages/runtime/src/registry.tsx
```diff
@@ -49,18 +50,25 @@ export type ComponentImplementation<T = any> = React.FC<
...
     subscribeMethods: SubscribeMethods;
     slotsMap: SlotsMap | undefined;
+    style?: CSSProperties;
+    data?: Record<string, any>;
   }
 >;
 
+export type TraitResult = {
+  props: {
+    data?: unknown;
```
Yuyz0112, line 60: Looks like we need to improve the spec so the trait can declare what kind of props it will return to the component?

packages/runtime/src/registry.tsx
tanbowensg: Argree, only one style is not enough.

packages/runtime/src/traits/core/fetch.tsx
tanbowensg: ok, I will fix it.

packages/runtime/src/traits/core/state.tsx
tanbowensg: Yes. I should add a lint rule for unused vairable.

packages/runtime/src/registry.tsx
tanbowensg: Yes we can.And the component need declare what kind of props it can deal with. But I think the trait result interface may change. We can do it after the interface become stable.

packages/runtime/src/traits/core/fetch.tsx
tanbowensg: I find it more complex than I thoughts. The best way to fix this problem is a memo mechanism which can detect change of properties, so that trait can decide whether to resend request. Maybe we can put this feature in next milestore.

Yuyz0112 (approved):

PR #51, opened 2021-09-03
event trait support any event name

tanbowensg:
event trait support any event name, not only click


Yuyz0112:
LGTM. 
And in the future, I think we can improve the type system to describe the events a component will emit.

Yuyz0112 (approved):

PR #52, opened 2021-09-05
Implement the Chakra UI Image in the runtime package.

chu-bin:
Implement the Chakra UI Image in the runtime package.

### DIFF
--- a/packages/runtime/example/image-component/index.html
+++ b/packages/runtime/example/image-component/index.html
@@ -1,0 +1,259 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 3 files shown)


packages/runtime/src/traits/core/event.tsx
```diff
@@ -55,6 +55,11 @@ const useEventTrait: TraitImplementation<{
           name: 'click',
         });
       },
+      onLoad() {
+        eventHandler({
+          name: 'onLoad',
+        });
+      },
```
tanbowensg, line 62: The lastest main branch has supported cutsom dynamic event names. So now we don't need hardcode here.

tanbowensg (changes requested):

packages/runtime/src/traits/core/event.tsx
chu-bin: got it

tanbowensg (approved):

PR #53, opened 2021-09-06
implement List component

tanbowensg:
# How does List work

1. accept listData property as data source.
2. eval template for each listItem.
3. use ImplWrapper to wrap every component in template, so template can be handled as normal component.
4. render template

# Custom Template

The user can pass the template content through the `template` property. The format of the template is the same as other components.

But State Trait should not be used in the template. Because each component in the template generates an ID dynamically, this dynamic ID cannot be accessed in the Schema, so even if the template has its own state, the state cannot be accessed in the Schema.

# Access ListItem

Users can access the value of listItem in the template through `$listItem` in the expression.


packages/runtime/src/util-methods.ts
```diff
@@ -15,3 +16,20 @@ export function mountUtilMethods() {
     }
   });
 }
+
+// parse component Type
+export function parseType(v: string) {
```
Yuyz0112, line 21: This file was designed for 'util methods can be invoked in the expression', like `{{ alert('hi') }}`, sorry for the misleading naming.

So I think parseType is a util function for this package?

packages/runtime/src/store.ts
```diff
@@ -3,43 +3,85 @@ import dayjs from 'dayjs';
...
 }
 
 export const stateStore = reactive<Record<string, any>>({});
 
-// TODO: use web worker
-const builtIn = {
-  dayjs,
-};
-function maskedEval(raw: string) {
+function parseExpression(str: string, parseListItem = false): ExpChunk[] {
```
Yuyz0112, line 26: I think we need some unit tests for this function

Yuyz0112:
@tanbowensg Could you write down some design decisions in this PR's description? Then it would be great for reviewers to get the whole picture of this feature, including 'why and why not'.

packages/runtime/src/store.ts
tanbowensg: OK, I have added unit test

packages/runtime/src/util-methods.ts
tanbowensg: Yes. Now I create a new folder for utils functions.

packages/runtime/jest.config.js
```diff
@@ -0,0 +1,194 @@
+/*
```
Yuyz0112, line 1: https://github.com/webzard-io/meta-ui/blob/main/packages/core/jest.config.js

We can use the general config as the base config.

packages/runtime/__tests__/expression.test.ts
```diff
@@ -0,0 +1,34 @@
+import { parseExpression } from '../src/store';
+
+describe('parseExpression function', () => {
+  it('can parse {{}} expression', () => {
+    expect(parseExpression('{{ value }}')).toMatchObject([
+      { expression: 'value', isDynamic: true },
+    ]);
+    expect(parseExpression('Hello, {{ value }}!')).toMatchObject([
+      { expression: 'Hello, ', isDynamic: false },
+      { expression: 'value', isDynamic: true },
```
Yuyz0112, line 10: the result has drop the last `!` character, is this expected?

Yuyz0112 (changes requested):

packages/runtime/__tests__/expression.test.ts
tanbowensg: It an expected bug. I fixed it.

packages/runtime/jest.config.js
tanbowensg: OK. And I found some problem with ts-jest, so I follow the documentation to add a babel.config.js instead of ts-jest.

Yuyz0112 (approved):

PR #57, opened 2021-09-08
feat: impl dialog component

jessie975:
impl dialog component & add a list example with dialog component

### DIFF
--- a/packages/runtime/example/dialog/index.html
+++ b/packages/runtime/example/dialog/index.html
@@ -29,21 +29,33 @@
               type: 'plain/v1/button',
               properties: {
                 text: {
-                  raw: `open`,
+                  raw: `**Open Dialog**`,
                   format: 'md',
                 },
                 colorScheme: 'red'
               },
               traits: [
                 {
+                  type: 'core/v1/slot',
+                  properties: {
+                    container: {
+                      id: 'root',
+                      slot: 'root',
+                    },
+                  },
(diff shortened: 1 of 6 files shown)


packages/runtime/example/dialog/index.html
```diff
@@ -0,0 +1,127 @@
...
+                    slot: 'root',
+                  },
+                },
+              }],
+            },
+            {
+              id: 'emailInput',
+              type: 'chakra_ui/v1/input',
+              properties: {
+                initValue: '{{ list.currentItem.email }}',
```
tanbowensg, line 82: It seems like there is not list in this file

packages/runtime/example/dialog/list.html
```diff
@@ -0,0 +1,671 @@
...
+                      },
+                      {
+                        event: 'confirmDialog',
+                        componentId: 'root',
+                        method: {
+                          name: 'setValue',
+                          parameters: {
+                            key: 'listData',
+                            value: `{{
+                              (function() {
```
tanbowensg, line 421: Although we support use function in expression, but it is not good pratice to write such complicated expression. It is hard to imagine user write long expression in mata-ui editor. We can divide it to 2 dialogs, edit and create dialog, to keep expression simple.

jessie975: ok

PR #61, opened 2021-09-13
implement table

tanbowensg:
# Changed Content
* add Table Component
* table CRUD with api
* add `onComplete` property for fetch trait and remove fetch `name` property
* fix component and trait excecution order. Ensure traits are excuted before components.
* initiate state when app starts

# Table Features
* pagination
* select and multiselect row
* sort by column
* text, image, button cell
## Table Properties
* `data`: Array. data source of table
* `marjorKey`:  majorKey of table items
* `rowsPerPage`
* `isMultiSelect`: enable multiselect
* `columns`: object that defines row 
```javascript
export const ColumnSchema = Type.Object({
  key: Type.String(),// key of item
  title: Type.String(),// title of td
  displayValue: Type.Optional(Type.String()), // use expression to format value
  type: TdTypeSchema,// text, image, button
  buttonConfig: Type.Object({
    text: Type.String(),
    events: Type.Array(
      Type.Object({
        componentId: Type.String(),
        method: Type.Object({
          name: Type.String(),
          parameters: Type.Any(),
        }),
      })
    ),
  }),
```

#Table State
* selectedItem: currently selected item(works in both single select mode and multi select mode).
* selectedItems: only works in multi select mode.


Yuyz0112 (approved):

Yuyz0112:
@tanbowensg LGTM, please resolve the conflicts then we can merge this.

PR #73, opened 2021-09-18
Complete Form Component

tanbowensg:
Form consists of 2 components: `Form` and `FormControl`.
# Form
## Properties
* `hideSubmit`: hide submit button
## State
* `isFormInvalid`
* `data`: current form data

# FormControl
## Properties
* `label`: input label
* `fieldName`: key of value in `data`
* `isRequired`
* `helperText`
## State
* `inputId`: id of input, for `Form` to send reset event to input.
* `fieldName`: for `Form` to build `data`
* `isInvalid`: is input invalid
* `value`: value of input
截屏2021-09-18 下午5 58 33

截屏2021-09-18 下午6 07 00


packages/runtime/src/App.tsx
```diff
@@ -135,9 +144,11 @@ export const ImplWrapper = React.forwardRef<
 
   // eval component properties
   useEffect(() => {
-    const { stop } = deepEval(c.properties, ({ result }) => {
-      setEvaledComponentProperties({ ...result });
+    const { result, stop } = deepEval(c.properties, ({ result: newResult }) => {
+      setEvaledComponentProperties({ ...newResult });
     });
+    // must keep this line, reason is the same as above
+    setEvaledComponentProperties({ ...result });
```
Yuyz0112, line 151: Good catch, also run into this issue in my branch.

Another issue that may not need to be changed in this PR is shall we remove the `deepEval` call to get the initial value of `evaledComponentProperties`?

Yuyz0112:
Really good job! There may be some design decisions that can be discussed in the future, e.g, the implementation of Form requires a lot of 'knowledge' of meta-ui, which can stop future users to contribute their components.

But currently, the PR is good enough and I would love to merge it.

Yuyz0112 (approved):

packages/runtime/src/App.tsx
Yuyz0112: never mind, just found it has already been removed.

Yuyz0112: update: it should be kept to make sure the component gets evaluated property at the first render.

PR #82, opened 2021-10-09
Add new component metadata

tanbowensg:
# New Metadata
* isDraggable: can this component be dragged in grid
* isResizable: can this component be resized in grid
* displayName: name displayed in component list
* icon: icon displayed in component list(haven't decided yet)
* defaultProperties: defaultProperties when component first be created

### DIFF
--- a/packages/core/src/metadata.ts
+++ b/packages/core/src/metadata.ts
@@ -8,5 +8,5 @@ export type ComponentMetadata = Metadata & {
   isResizable: boolean;
   displayName: string;
   icon?: string;
-  defaultProperties?: Record<string, unknown>;
+  defaultProperties: Record<string, unknown>;
 };
--- a/packages/core/src/metadata.ts
+++ b/packages/core/src/metadata.ts
@@ -8,5 +8,5 @@ export type ComponentMetadata = Metadata & {
   isResizable: boolean;
   displayName: string;
   icon?: string;
-  defaultProperties: Record<string, unknown>;
+  exampleProperties: Record<string, unknown>;
 };
--- a/packages/core/src/metadata.ts
+++ b/packages/core/src/metadata.ts
@@ -2,3 +2,11 @@ export type Metadata = {
   name: string;
   description?: string;
 };
+
+export type ComponentMetadata = Metadata & {
+  isDraggable: boolean;
+  isResizable: boolean;
+  displayName: string;
+  icon?: string;
+  defaultProperties?: Record<string, unknown>;
+};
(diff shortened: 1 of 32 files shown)


packages/core/src/metadata.ts
```diff
@@ -2,3 +2,11 @@ export type Metadata = {
   name: string;
   description?: string;
 };
+
+export type ComponentMetadata = Metadata & {
+  isDraggable: boolean;
```
Yuyz0112, line 7: which kind of component could not be dragged into a grid?

packages/core/src/metadata.ts
```diff
@@ -2,3 +2,11 @@ export type Metadata = {
...
+  isResizable: boolean;
+  displayName: string;
```
Yuyz0112, line 9: I think our version field `category/name` already provides a unique name for a component. So in the spec aspect, shall we tell end-users when and how to set displayName?

packages/editor/src/operations/OperationManager.ts
```diff
@@ -30,7 +30,8 @@ function genComponent(
 ): ApplicationComponent {
   const { version, name } = parseType(type);
   const cImpl = registry.getComponent(version, name);
-  const initProperties = parseTypeBox(cImpl.spec.properties as any);
+  const initProperties =
+    cImpl.metadata.defaultProperties || parseTypeBox(cImpl.spec.properties as any);
```
Yuyz0112, line 34: do we need a merge implementation so users can set partial defaultProperties?

packages/runtime/src/components/chakra-ui/Image.tsx
```diff
@@ -123,7 +123,17 @@ export default {
     version: 'chakra_ui/v1',
     metadata: {
       name: 'image',
+      displayName: 'Image',
       description: 'chakra_ui image',
+      isDraggable: true,
+      isResizable: true,
+      defaultProperties: {
```
Yuyz0112, line 130: Since this kind of default value looks like a set of subjective mock values and is mostly used in editor mode(I think?), is it possible to make the code/naming more concrete?

Like `exampleProperties or `editorInitProperties` or other smart namings. Because default properties sound like the best practice props in runtime and end-users can use the component really well without other props.

tanbowensg: Good point. exampleProperties will be better.

packages/core/src/metadata.ts
tanbowensg: In my opinion, the category/name is for programming, not for display. For example, GridLayout's name is `grid_layout`, but maybe we want display it as `Grid Layout` or something else. I think declare it in spec is more convinient than somewhere else.

Yuyz0112: fair enough

Issue #54, opened 2021-09-07
Update Spec 


tanbowensg:
# Component Spec
1. remove `acceptTrait`
2. add `styleSlots` and `events`. Both of them are simple string array.
```
 type Component = {
  version: string;
  kind: 'Component';
  metadata: Metadata;
  spec: ComponentSpec;
};

type ComponentSpec = {
  properties: JSONSchema;
  state: JSONSchema;
  methods: MethodSchema[];
  styleSlots: string[];
  slots: string[];
  events: string[]
};
```

# Trait Spec
1. add `returnProps`. `returnProps` are keys of the object returned by trait. For now, there are only 2 keys: `styles` and `callbackMap`. Component can only deal with these 2 props.

```
 type Trait = {
  version: string;
  kind: 'Trait';
  metadata: Metadata;
  spec: TraitSpec;
};

type TraitSpec = {
  properties: JSONSchema;
  state: JSONSchema;
  methods: MethodSchema[];
  returnProps: string[]// styles, callbackMap...
};
```

Yuyz0112:
Another proposal: update `ComponentSpec.properties` and `TraitSpec.properties` from `JSONSchema[]` to `JSONSchema`.

Yuyz0112:
@tanbowensg I've updated the proposal with an example. If this looks reasonable to you, I'm going to send a PR.

tanbowensg:
Ok.LGTM. The new spec looks simpler and more precise.

Yuyz0112:
@tanbowensg I think the original proposal of this issue still needs an update?

PR #84, opened 2021-10-12
Update Component Spec

tanbowensg:
add some fields for component as described in #54 

  styleSlots: string[];
  slots: string[];
  events: string[];

and add `exampleSize` in component metadata

### DIFF
--- a/packages/runtime/example/box/index.html
+++ b/packages/runtime/example/box/index.html
@@ -40,14 +40,13 @@
                   properties: {
                     events: [
                       {
-                        event: 'click',
+                        event: 'onClick',
                         componentId: 'test_btn',
                         method: {
                           name: 'setValue',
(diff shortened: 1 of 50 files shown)


packages/runtime/example/box/index.html
```diff
@@ -40,14 +40,13 @@
                   properties: {
                     events: [
                       {
-                        event: 'click',
+                        event: 'onClick',
```
Yuyz0112, line 43: I think the `onEvent` naming may not be very friendly to editor users. For example, when they configure events handler in the GUI, `click`, `submit` looks more straightforward.

tanbowensg:
Can this pr be merged? @Yuyz0112

Yuyz0112 (approved):

PR #92, opened 2021-10-18
introduce CodeMirror to implement code editors

Yuyz0112:
According to other open-source projects' implementation, it looks like CodeMirror has the best choice to customize a low-code editor.

In this PR, we introduce CodeMirror and set up a basic JSON viewer for our store state.


Yuyz0112:
image

PR #93, opened 2021-10-19
feat: modify  nested properties & traits

chu-bin:
## Component Form

- can edit nested properties
- all component properties add into component form, example properties will be fill in component form
- can modify and undo modify component id
- can edit nested trait

### DIFF
--- a/packages/editor/src/operations/AppModelManager.ts
+++ b/packages/editor/src/operations/AppModelManager.ts
@@ -6,10 +6,12 @@ import {
   RemoveComponentOperation,
   ModifyComponentPropertyOperation,
   ModifyTraitPropertyOperation,
+  ModifyComponentIdOperation,
 } from './Operations';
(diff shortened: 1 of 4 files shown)


packages/editor/src/operations/AppModelManager.ts
```diff
@@ -22,6 +24,16 @@ function genSlotTrait(parentId: string, slot: string): ComponentTrait {
     },
   };
 }
+// TODO JSONSchema7Type
+const genInitProps = (init: Record<string, never>, properties: any, fullKey?: string) => {
```
tanbowensg, line 28: there is a similar tool function `parseTypeBox` in runtime. Maybe we can reuse it.

chu-bin: Thanks, I‘m already used this function and moved it to before rendering component form

PR #94, opened 2021-10-20
edit event trait

tanbowensg:
* can add/remove/edit event trait


packages/editor/src/components/ComponentForm/ComponentForm.tsx
```diff
@@ -74,43 +70,74 @@ export const ComponentForm: React.FC<Props> = props => {
...
-
-    return (
-      <form key={t.type}>
-        <strong>{t.type}</strong>
-        {traitForm}
-      </form>
-    );
   });
 
+  const eventHandlers = selectedComponent.traits.find(t => t.type === 'core/v1/event')
```
Yuyz0112, line 90: Maybe we can pass selectedComponent to `EventTraitForm`, and let it handle this inside the component.

This will help others define their own trait/component form without change the `ComponentForm` code.

Yuyz0112, line 90: Maybe we can pass `selectedComponent` to `EventTraitForm`, and let it handle this inside the component.

This will help others define their own trait/component form without changing the `ComponentForm` code.

PR #100, opened 2021-11-01
Some small features

tanbowensg:
* add some components: link, divider
* use string as style trait property
* change tabs slot to static
* update vite version


Yuyz0112 (approved):
LGTM. Look like my previous PR break the jest CI, going to send a PR to fix it.

Issue #102, opened 2021-11-02
Module Mechanism


tanbowensg:
# Use cases
* You build a complicated form by meta-ui and want to reuse it in multiple meta-ui application.
* You build a fancy listItemView and want to use it in list.

# module vs component
Module is made up of meta-ui component. Component is created in react.

# Features of module
Basicly, module can be used like a component. Module can receive properties and expose state & events.

# Tasks
- [x] define spec and schema
- [x] render module as item of table td and listItem
- [x] render module individually
- [x] create module in editor
- [x] edit module in editor

PR #112, opened 2021-11-04
Feat/module

tanbowensg:
# Module Description
#102 

# Usage

```
{
  id: 'form',
  type: 'core/v1/moduleContainer',
  properties: {
    id: 'littleFormModule',
    type: 'core/v1/littleForm',
    properties: {
      btnText: 'Submit!',
    },
    handlers: [
      {
        type: 'onSubmit',
        componentId: '$utils',
        method: {
          name: 'alert',
          parameters:
            'Submit: {{ littleFormModule ? littleFormModule.value : 0 }}！',
        },
        wait: {},
        disabled: false,
      },
    ],
  },
  traits: [],
}
```

# Remain Problem
Cannot use module exposed state in module handler when using module as list template. Because current expression doest not support something like this `{{ moduleName{{$listItem.id}}.stateValue }}`


packages/runtime/example/module/index.html
```diff
@@ -0,0 +1,137 @@
...
+            btnText: Type.String(),
+          },
+          events: ['onSubmit'],
+          stateMap: {
+            value: '{{$moduleId}}__input.value',
+          },
+        },
+        components: [
+          {
+            id: '{{$moduleId}}__hstack',
```
Yuyz0112, line 34: Adding `moduleId` as a prefix to avoid components id conflictions looks reasonable, but writing the boilerplate many times maybe a little boring for end users.

Not sure whether we should add this automatically and let the user only write `id: 'hstack'` or not, but it's out of this PR's scope so we can discuss about it in the coming weekly meeting.

Yuyz0112 (approved):

PR #114, opened 2021-11-06
 migrate HTML examples to JSON

Yuyz0112:
# What's included in this PR

1. Implement a vite plugin, which will scan the new examples folder and compose all the JSON schemas into one JSON file.
2. Migrate all the HTML examples to JSON, which lets us re-use examples across packages.

# Changes to developers

Instead of visiting multiple URLs, now runtime package developers can visit `localhost:3000` to visit all the examples.

The format of an example is:

```json
{
  "modules": [] // if you have a custom module, can be declared here. custom traits and components should be implemented in the same way.
  "app": { ... } // the app schema
}
```

And there are some limitations of the new JSON format examples:

1. Multiline strings are not supported by JSON, so we need `\n` to format multiline strings.
2. We could not use typebox in JSON, so we need to write raw JSON schema as an alternative.

# The new dev playground

image
image


Yuyz0112:
Some examples seem broken, but not sure it's caused by this migration.

PR #118, opened 2021-11-08
Introduce JSON schema form as the default component properties form

Yuyz0112:
issue: https://github.com/webzard-io/meta-ui/issues/109

## Design

The design of our JSON schema form is largely inspired by [rjsf](https://github.com/rjsf-team/react-jsonschema-form). In the form, there are two main concepts: `schema`(the static JSON schema) and `formData`(the dynamic runtime form value).

The architecture:
image

The form will read from the schema and pass the schema to the `SchemaField`. `SchemaField` will iterate the passed-in schema and choose the right field to handle it. For example, if the `SchemaField` gets a schema with type `object`, it will render an `ObjectField`.

We will have the following fields:

[x] `ObjectField`
[x] `StringField`
[x] `ArrayField`
[x] `BooleanField`
[x] `MultiSchemaField`, for `anyOf` and `oneOf` schemas.
[x] `NumberField`
[x] `NullField`
[x] `UnsupportedField`

Fields, for example, the `ObjectField`, can use `SchemaField` inside itself to do the recursive rendering.

The `SchemaField` will wrap a `Template` around every field, which is a best practice for form controls.

image

The `title`, `description`, `help` properties will be read from the schema, and `errors` will be maintained by the form.

We can allow customizing the template by passing a `customTemplate` component to the `SchemaField`.

Inside the fields, it will use widgets to do the rendering. For example, if a field has a type of `string` in the schema, it will use the `Input` widget as its renderer. But if the schema indicates the field can only accept two enum strings, then the widget should be changed into a `Select` component.
Otherwise, if there is more information from the schema, it can choose widgets with better UX.

### DIFF
--- a/packages/core/src/component.ts
+++ b/packages/core/src/component.ts
@@ -1,4 +1,4 @@
-import { JSONSchema7, JSONSchema7Object } from 'json-schema';
+import { JSONSchema7 } from 'json-schema';
 import { parseVersion } from './version';
 import { ComponentMetadata } from './metadata';
 import { MethodSchema } from './method';
@@ -14,7 +14,7 @@ export type Component = {
 };
 
 type ComponentSpec = {
-  properties: JSONSchema7Object;
+  properties: JSONSchema7;
   state: JSONSchema7;
   methods: MethodSchema[];
   styleSlots: string[];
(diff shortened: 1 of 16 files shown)


Yuyz0112:
Also improve the `parseTypeBox` utils, which support generating enum and union schema values.

tanbowensg (approved):
Very Good

PR #122, opened 2021-11-10
add multiSelecteTree & treeSelect & some other changes

tanbowensg:


packages/runtime/src/components/_internal/Text.tsx
```diff
@@ -1,5 +1,6 @@
 import { css } from '@emotion/react';
 import ReactMarkdown from 'react-markdown';
+import { Text as BaseText } from '@chakra-ui/react';
```
Yuyz0112, line 3: I don't think we should dependent chakra-ui in internal/core components

packages/runtime/src/components/antd/TreeSelect.tsx
```diff
@@ -0,0 +1,174 @@
+import { useState, useEffect } from 'react';
+import { createComponent } from '@meta-ui/core';
+import { Static, Type } from '@sinclair/typebox';
+import { uniq } from 'lodash';
```
Yuyz0112, line 4: please use `lodash-es`

Yuyz0112 (changes requested):

Yuyz0112 (approved):

PR #136, opened 2021-11-24
support unmount component when hidden

Yuyz0112:
Currently, we only support hidden a component visually, which is not good for performance and miss the control of the component life cycle.

In this PR, I've added a new prop to the hidden trait, which controls the hidden is visually or logically.

The short of this implementation is it could not re-use the `props` field in `TraitResult`, so I added a new field `unmount` to let the `ImplWrapper` know what to do. This makes the `TraitResult` type looks casual, any suggestion on the type will be appreciated.


tanbowensg:
Since `unmount` is only used by `ImplWrappr`, so I think it is fine to be an individual field in `TraitResult` for now. Maybe we will have more fields like this in the future.

tanbowensg (approved):

Issue #106, opened 2021-11-02
Component Id reference dose not change after component id changed.


tanbowensg:
# Places need to be modify after component id changes
* slot trait
* event trait
* expressions used this component Id

tanbowensg:
Missing situation:
- [ ]  id of method target
- [ ] id in expression

Issue #79, opened 2021-09-28
edit schema Operations


tanbowensg:
* support undo/redo
* add/remove/edit component/trait
* move component

Yuyz0112:
For undo/redo, some references:

1. JSON patch: http://jsonpatch.com/
2. CRDT implemented in JS: https://github.com/automerge/automerge

PR #139, opened 2021-11-25
editor operation

Sczlog:
## editor operation

define an operation system for sunmao-ui's editor #79 #106 

define every user interaction as an operation, and an operation can be defined as such an interface:

```typescript
export interface IOperation<TContext = any> extends IUndoRedo {
  /**
   * the context of operation, provide any information an operation needed
   */
  context: TContext;
  /**
   * infer the type of operation, leaf or branch
   */
  type: Symbol;
  do(prev: Application): Application;
  redo(prev: Application): Application;
  undo(prev: Application): Application;
}
```

context contain information the operation needed.

do is called for the first time the operation is applied, input the current application state and it will output the next application state by immer produce.
redo/undo are two inverse method.

type is used to infer the type of operation, we have two kind of operation:

1. leaf operation:
   leaf operatio is an atomic operation, which mean it will not has side effect to the part irrelevant to it.
   and it is the operation which will modify json schema directly by its do/redo/undo method, do and undo are inverse method to its opposite.

2. branch operation:
   branch operation is a group of operations(branch or leaf), the branch operation will not directly modify json schema, but by its children.
   branch operation is also an undoredo manager, which maintain a doubly linked list as operation stack. it will traverse the list foward and reverse to do/redo/undo operation.

```typescript
export interface IUndoRedoManager extends IUndoRedo {
  operationStack: OperationList;
}
export class OperationList<TOperation extends IOperation = IOperation> {
  private readonly head = new ListNode<null, never, TOperation | null>(null);
  private _cursor: ListNode<TOperation | null>;
  public get cursor(): ListNode<TOperation | null> {}
  constructor() {}
  insert(value: TOperation) {}
  moveNext() {}
  movePrev() {}
  reduce<TData>(
    cb: (prev: TData, node: TOperation) => TData,
    initialData: TData
  ): TData {}
  reduceRight<TData>(
    cb: (prev: TData, node: TOperation) => TData,
    initialData: TData
  ): TData {}
}
```

The top level of the operation system is also an IUndoRedoManager, but not like branch operation, it will not traverse entire list but use a cursor from the operation stack to identified the state of the application is based on which operation, and call redo/undo method by moving the cursor by one step, a dummy null node is represent the initial state.

image


packages/editor/src/operations/leaf/component/freeOperation.ts
```diff
@@ -0,0 +1,38 @@
+import { Application, ApplicationComponent } from '@sunmao-ui/core';
+import produce from 'immer';
+import { BaseLeafOperation } from '../../type';
+
+export type FreeComponentOperationContext = {
+  componentId: string;
+};
+
+export class FreeComponentOperation extends BaseLeafOperation<FreeComponentOperationContext> {
```
tanbowensg, line 9: Fee -> Remove

packages/editor/src/operations/AppModelManager.ts
```diff
@@ -1,301 +1,73 @@
...
+      return;
+    }
+    try {
+      this.operationStack.moveNext();
+    } catch {
+      console.warn('cannot redo as cannot move to next cursor', this.operationStack);
+      return;
+    }
+    const newApp = this.operationStack.cursor?.val?.do(this._app);
+    // console.log('redo', this.operationStack.cursor?.val);
```
tanbowensg, line 37: remove console.log

tanbowensg (changes requested):

tanbowensg (approved):

PR #142, opened 2021-11-29
fix adjust order operation

Sczlog:
1. fix adjust order operation
rewrite logic to adjust order of operation, just switch two element in the position.
2. add unit test for the operation.


tanbowensg (approved):
LGTM

packages/editor/__tests__/operation/__snapshots__/adjustComponentOrder.spec.ts.snap
```diff
@@ -0,0 +1,2725 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
Yuyz0112, line 1: @Sczlog  @tanbowensg 

I don't think the current size of snapshots is reviewable. Could we use a tiny but comprehensive fixture and only use small inline snapshots when needed. Otherwise, we should write other assertions to test the expected behavior.

PR #143, opened 2021-11-29
Support edit module in editor

tanbowensg:
# Features
* can load and edit module schema in editor (not spec yet)
# Refactor
Because now editor need to edit components of `Application` and `Module`. So I change the editor parameter from `Application` to `components`. It is ok beacuse the schema of components of `Application` and `Module` are the same. Maybe Runtime can just accept `components` in the future.


Yuyz0112:
`Application` can derive more metadata in the future which makes it looks more different from `Module`. But the editor accepting `components` looks reasonable to me.

Yuyz0112:
CI failed

PR #167, opened 2021-12-14
introduce widgets and impl module widget

Yuyz0112:
With this PR, component developers can define the editor widget of certain fields like this:

```js
module: Type.Object(RuntimeModuleSchema.properties, {
  widget: 'module',
})
```

In the future, we can support registering custom widgets to an editor, then developers can define an editor widget along with the component implementation.

### DIFF
--- a/packages/editor/src/components/ComponentForm/ComponentForm.tsx
+++ b/packages/editor/src/components/ComponentForm/ComponentForm.tsx
@@ -134,6 +134,7 @@ export const ComponentForm: React.FC<Props> = observer(props => {
                 })
               );
             }}
+            registry={registry}
           />
         </VStack>
       </VStack>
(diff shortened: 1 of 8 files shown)


Yuyz0112:
@tanbowensg any suggestions?

tanbowensg:
Looks very useful. Can we implement trait forms as widget ? like `EventHandlerForm` ?

Yuyz0112:
@tanbowensg Sure. We can change the props schema of event trait to something like this:

```js
const PropsSchema = Type.Object(
  {
    handlers: Type.Array(EventHandlerSchema),
  },
  {
    wdiget: 'event-trait',
  }
);
```

And registering the event-trait widget.

tanbowensg (approved):
LGTM

PR #168, opened 2021-12-15
improve the Slot component

Yuyz0112:
1. bypass props to children
2. render null when slot not set

### DIFF
--- a/packages/runtime/src/components/chakra-ui/Tabs.tsx
+++ b/packages/runtime/src/components/chakra-ui/Tabs.tsx
@@ -36,7 +36,7 @@ const Tabs: ComponentImplementation<Static<typeof PropsSchema>> = ({
         ))}
       </TabList>
       <TabPanels>
-        {getSlots(slotsMap, 'content').map((content, idx) => {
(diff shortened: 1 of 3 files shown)


packages/runtime/src/components/chakra-ui/Tabs.tsx
```diff
@@ -36,7 +36,7 @@ const Tabs: ComponentImplementation<Static<typeof PropsSchema>> = ({
         ))}
       </TabList>
       <TabPanels>
-        {getSlots(slotsMap, 'content').map((content, idx) => {
+        {getSlots(slotsMap, 'content', {}).map((content, idx) => {
```
Yuyz0112, line 39: Not sure whether these `getSlots` calls can be replaced by `<Slot />` after this patch.

tanbowensg: Maybe not. In tabs, every components in slot should be wrapped with panel. Only by `getSlots` can we get the array of components in slot.

tanbowensg (approved):

PR #174, opened 2021-12-20
refactor style trait

tanbowensg:
1. style trait accept array
2. refactor style trait form

### DIFF
--- a/packages/runtime/src/traits/core/style.tsx
+++ b/packages/runtime/src/traits/core/style.tsx
@@ -2,22 +2,25 @@ import { createTrait } from '@sunmao-ui/core';
 import { Static, Type } from '@sinclair/typebox';
 import { TraitImplementation } from '../../types/RuntimeSchema';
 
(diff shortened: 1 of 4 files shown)


packages/runtime/src/traits/core/style.tsx
```diff
@@ -2,22 +2,25 @@ import { createTrait } from '@sunmao-ui/core';
...
-      },
+      customStyle,
     },
   };
 };
 
 const PropsSchema = Type.Object({
-  styleSlot: Type.String(),
-  style: Type.String(),
+  styles: Type.Array(
```
Yuyz0112, line 18: Do any examples need migration?

Yuyz0112:
**BREAKING CHANGE**

packages/runtime/src/traits/core/style.tsx
tanbowensg: Yes, I have already updated examples.

tanbowensg: Well, actually there is no style trait example, only some tests.

Issue #135, opened 2021-11-23
move chakra-ui components to a standalone package


Yuyz0112:
`@sunmao-ui/runtime` should not be dependent on `chakra-ui`

Yuyz0112:
@tanbowensg I prefer to work on this in these two days. Then CloudTower can use sunmao-ui runtime without bundle chakra-ui.

PR #175, opened 2021-12-21
Lib chakra-ui

Yuyz0112:
#135 

---

- Export the standard `SunmaoLib` type to help library builder.
- Move Chakra-UI components to a new package `sunmao-chakra-ui-lib`.

The runtime package's example playground is using the new lib package. The editor package function has not been verified manually.

In the future, we can move the lib package into a standalone repo.

**TODO:**

1. `$utils.toast` is broken because we are not able to register util methods at this moment.
2. We are exporting the `watch` method because some chakra-ui lib components depend on it, I hope this can be removed in the future.


tanbowensg (approved):
LGTM. I have created issues for todos.

PR #179, opened 2021-12-22
upgrade tsup and compile to es5

Yuyz0112:


Yuyz0112:
tsup use swc under the hood instead of babel, which looks fast and easy to use.

tanbowensg (approved):
LGTM

PR #180, opened 2021-12-25
Live in a safer world:  type-safe component for sunmao-ui

Yuyz0112:
sunmao-ui's type system provides two kinds of type safety: type-safe application and type-safe component, this PR only discusses the latter one.

# The design

When a user develops a sunmao-ui component, they may need type checking to their component's code, which guarantees the usage of sunmao-ui's APIs.

> In the current version, sunmao-ui's runtime is React, so we mainly focus on providing type safety in React system.

## The usage of sunmao-ui APIs

sunmao-ui injects a branch of APIs into a component as props, which is a common pattern of React-based component development. For example, a component will get a prop called `customStyle`, which contains the styles defined by style traits. Since a component's spec defined the name of the style slots, the `customStyle` should know only the style slots' names are the valid key of this object.

The full list of APIs we are going to provide type checking:

- `mergeState`
- `subscribeMethods`
- `callbackMap`
- `customStyle`
- slot
- other component props

## Build a bridge between JSON-schema types and typescript types

As we mentioned above, sunmao-ui already lets components define a JSON-schema-based spec to let the runtime knows how to interact with components. So our design will leverage the power of typescript to infer as much information as possible from the JSON-schema types, which helps users do not to need to define the same types again in two systems.

This is mainly archived by using typescript's generic system and [typebox](https://github.com/sinclairzx81/typebox)'s typescript-based JSON-schema builder.

Finally, we got an API like this: `implementRuntimeComponent(spec)(component-implementation)`. The component-implementation is a fully-typed React component.

## BREAKING CHANGES

We changed the spec of `methods` to help typescript infer types.

# TODO

- [ ] migrate all components to the new API
- [ ] remove all `TODO: (type-safe)` marks


tanbowensg (approved):

tanbowensg:
I have updated all the components.
But `slotsMap` cannot be removed now becuase `form` and `tabs` rely on it. And `form` component is broken after slot updated.
I will refactor slot and  solve these problems in next PR.

PR #182, opened 2021-12-28
Add model for component and trait

tanbowensg:
# Features
1. add model
3. refactor operations with model
3. refactor validator with model
4. add unit tests for validator and model

Although there are more works to do, while this pr is too large. So I put features below to the future.

# Todo
1. refactor operationManager with model
2. refactor UI with model
3. parse dependencies in expression
4. implement module model


packages/editor/__tests__/model/appModel.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+describe('AppModel test', () => {
+  const appModel = new AppModel(AppSchema.spec.components);
+  describe('resolve tree', () => {
+    it('resolve Tree corectlly', () => {
+      expect(appModel.allComponents.length).toBe(10);
+      expect(appModel.topComponents.length).toBe(2);
+    });
+
+    it('components order', () => {
+      expect(appModel.allComponents[0].id).toBe('hstack1');
```
Yuyz0112, line 14: maybe `expect(appModel.allComponents.map(c => c.id)).toMatchInlineSnapshot()`

packages/editor/src/AppModel/AppModel.ts
```diff
@@ -0,0 +1,133 @@
...
+    delete this.componentMap[componentId];
+    if (comp.parentSlot && comp.parent) {
+      const children = comp.parent.children[comp.parentSlot];
+      comp.parent.children[comp.parentSlot] = children.filter(c => c !== comp);
+    } else {
+      this.topComponents.splice(this.topComponents.indexOf(comp), 1);
+    }
+  }
+
+  _bindComponentToModel(component: IComponentModel) {
```
Yuyz0112, line 78: use private instead of _ prefix?

packages/editor/src/validator/rules/TraitRules.ts
```diff
@@ -122,11 +107,9 @@ class EventHandlerValidatorRule implements TraitValidatorRule {
         return;
       }
 
-      if (
-        methodSchema?.parameters &&
-        !ajv.validate(methodSchema.parameters, parameters)
-      ) {
+      if (method.parameters && !ajv.validate(method.parameters, parameters)) {
         ajv.errors!.forEach(error => {
+          JSON;
```
Yuyz0112, line 112: what's this

Yuyz0112 (changes requested):

packages/editor/src/validator/rules/TraitRules.ts
tanbowensg: fixed

packages/editor/src/AppModel/AppModel.ts
tanbowensg: Because `ComponentModel` need to call this method, it cannot be private. But it should not be called outside, so I add `_` before it.

packages/editor/__tests__/model/appModel.spec.ts
tanbowensg: I use `toEqual` instead. I think it is clearer.

PR #196, opened 2022-01-04
manage editor styles in a single CSS

Yuyz0112:
Some frameworks like Next.js does not like global CSS from node
modules, so we are going to bundle CSS into a single file with
postcss.

ref: https://github.com/vercel/next.js/discussions/27953


Yuyz0112:
test slack

tanbowensg (approved):

PR #200, opened 2022-01-04
multiple slot drag handling

Sczlog:
1. add dragover event handler to make drop event handled correctly.
2. get correct drag over component id when drag between two draggable component

updated in 2022/01/05:
1. add support for multiple slots dnd support

### DIFF
--- a/packages/editor/src/EditorStore.ts
+++ b/packages/editor/src/EditorStore.ts
@@ -16,7 +16,7 @@ class EditorStore {
   // currentEditingComponents, it could be app's or module's components
   selectedComponentId = '';
   hoverComponentId = '';
-  dragIdStack: string[] = [];
+  dragOverId: string = '';
   // current editor editing target(app or module)
   currentEditingTarget: EditingTarget = {
     kind: 'app',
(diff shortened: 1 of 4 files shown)


packages/editor/src/EditorStore.ts
```diff
@@ -43,23 +43,22 @@ class EditorStore {
   }
 
   get dragOverComponentId() {
```
tanbowensg, line 45: can remove

packages/editor/src/components/ComponentWrapper.tsx
```diff
@@ -20,16 +20,23 @@ export const ComponentWrapper: ComponentWrapperType = observer(props => {
...
 
-  const isDroppable = slots.length > 0;
+  useEffect(() => {
+    if (isDroppable) {
+      const listener = (e: Event) => e.preventDefault();
+      document.addEventListener('dragover', listener);
+      return () => {
+        document.removeEventListener('dragover', listener);
+      };
+    }
```
tanbowensg, line 38: why preventDefault document dragover?

Sczlog: it should on the element, modified

PR #216, opened 2022-01-19
fix: set slot to different types depend on slot children's num

MrWindlike:
when only one child in slot, set slot to ReactElement instead of array


tanbowensg:
The type of `ComponentImplProps` also need change.

MrWindlike:
> The type of `ComponentImplProps` also need change.

done.

tanbowensg:
Fail tests. Need fix.

tanbowensg (approved):

PR #223, opened 2022-01-25
refactor operationManager with appModel

tanbowensg:
I replace all the `ComponentSchema[]` to `AppModel` in every operations so that operation don't have convert  `ComponentSchema[]` to `AppModel` every time.


packages/editor/src/operations/branch/createComponentBranchOperation.ts
```diff
@@ -17,10 +17,10 @@ export type CreateComponentBranchOperationContext = {
 };
 
 export class CreateComponentBranchOperation extends BaseBranchOperation<CreateComponentBranchOperationContext> {
-  do(prev: ComponentSchema[]): ComponentSchema[] {
+  do(prev: AppModel): AppModel {
     // gen component id
     if (!this.context.componentId) {
-      this.context.componentId = genId(this.context.componentType, prev);
+      this.context.componentId = prev.genId(this.context.componentType as ComponentType);
```
Sczlog, line 23: should we narrow the type of context.componentType to ComponentType from string?

Sczlog:
LGTM

packages/editor/src/AppModel/ComponentModel.ts
```diff
@@ -170,9 +170,12 @@ export class ComponentModel implements IComponentModel {
   removeChild(child: IComponentModel) {
     const slotChildren = this.children[child.parentSlot!];
     if (slotChildren) {
-      slotChildren.splice(slotChildren.indexOf(child), 1);
-      child._isDirty = true;
-      this._isDirty = true;
+      const index = slotChildren.indexOf(child)
+      if (index > -1) {
```
Yuyz0112, line 174: when index will be -1?

tanbowensg: when `child` is not child of this component just having the same slotName.

Yuyz0112 (approved):

PR #229, opened 2022-01-26
fix(editor): make the form update correctly

MrWindlike:
### What is the PR about?
Fix the error with removing the item from array type property.

### What is the code about?
- Change the properties update logic
  Before, it will call `FieldModel.prototype.update` when remove a property of array or object. But, it just update the value by keys instead of removing the key. So now I change to put the existed keys and values to new array or object, then the others will be removed.
- Make the form can be controlled
  When remove the items of array, the sort of items are changed but the form would not be updated.


tanbowensg:
have conflict

tanbowensg (approved):

PR #231, opened 2022-01-27
fix: Fix the bug that the component will be deleted when editing in the Input

xzdry:


packages/editor/src/components/KeyboardEventWrapper.tsx
```diff
@@ -104,15 +108,17 @@ export const KeyboardEventWrapper: React.FC<Props> = ({
...
-      id="keyboard-event-wrapper"
-      className={style}
-      width="100%"
-      height="100%"
-      onKeyDown={onKeyDown}
-      tabIndex={-1}
-    >
-      {children}
-    </Box>
+    <>
```
tanbowensg, line 111: why add this?

packages/editor/src/components/KeyboardEventWrapper.tsx
```diff
@@ -32,6 +32,10 @@ export const KeyboardEventWrapper: React.FC<Props> = ({
     switch (e.key) {
       case 'Delete':
       case 'Backspace':
+        // if `selectionStart` available,think it is textarea or text input
+        if ((e.target as any).selectionStart !== undefined) {
```
Yuyz0112, line 36: looks like checking the target's tag name is a more straightforward and easy-to-understand way of doing this?

xzdry: seems better，i'll check it out

tanbowensg: Checking tag name cannot detect whether the input is focused.

tanbowensg (approved):

packages/editor/src/components/KeyboardEventWrapper.tsx
xzdry: > Checking tag name cannot detect whether the input is focused.

It should be possible, the result of my test is that the tagname is `INPUT` only in the focus state

PR #233, opened 2022-01-27
fix(editor): make the warning area update immediately

MrWindlike:


packages/editor/src/validator/rules/TraitRules.ts
```diff
@@ -39,7 +39,8 @@ class EventHandlerValidatorRule implements TraitValidatorRule {
         });
       }
 
-      if (isExpression(targetId)) {
+      // TODO: util methods has no method schema to check the parameters, so now temporally skip validation
+      if (isExpression(targetId) || targetId === GLOBAL_UTILS_ID) {
```
Yuyz0112, line 43: shall we validate the method without schema as `any`? Looks like currently, we allow users register method with a loose schema.

MrWindlike: Yes. Editor may support auto complete the params of method, and it depends on the specific schema.

tanbowensg: We may add spec for utilMethods in the future.

tanbowensg (approved):
We may add utilMethods spec in the future

PR #237, opened 2022-01-28
fix(editor): fix some errors and add test cases

MrWindlike:
- fix production error by tsup
- fix that can't find the component by id after change its id


packages/editor/src/AppModel/AppModel.ts
```diff
@@ -80,6 +80,11 @@ export class AppModel implements IAppModel {
     }
   }
 
+  changeComponentMapId(oldId: ComponentId, newId: ComponentId) {
```
tanbowensg, line 83: Since this function is always called after `Component.changeId`, maybe we can move this logic to `Component.changeId`?

MrWindlike: May should put `ComponentModel.changeId` into `AppModel.changeComponentMapId`? Because `AppModel` is the controller of `ComponentModel`.

tanbowensg: No. `ChangeId` is more reasonable and natural to be a method of `ComponentModel` than of `AppModel`.

MrWindlike: OK.

tanbowensg (approved):

PR #238, opened 2022-01-28
feat: support top level components reorder

xzdry:


tanbowensg:
add a unit test to cover this bug

tanbowensg (approved):

PR #243, opened 2022-01-29
feat: change selected componet will automatically jump to the inspect…

xzdry:


packages/editor/src/services/EditorStore.ts
```diff
@@ -180,4 +192,12 @@ export class EditorStore {
   setLastSavedComponentsVersion = (val: number) => {
     this.lastSavedComponentsVersion = val;
   };
+
+  setLeftTabIdx = (val: number) => {
```
Yuyz0112, line 196: how about some more concrete methods, e.g, `activeExplorer`, `activeInspect`

packages/editor/src/services/EditorStore.ts
```diff
@@ -19,6 +19,10 @@ export class EditorStore {
   _selectedComponentId = '';
   _hoverComponentId = '';
   _dragOverComponentId = '';
+  // default left tab: Explorer
+  _leftTabIdx = 0;
```
Yuyz0112, line 23: we can use an enum to represent indexes in a more semantic way

```ts
enum LeftTab {
  EXPLORER,
  UI_TREE,
  STATE
}
```

Yuyz0112 (changes requested):

packages/editor/src/components/ComponentWrapper.tsx
```diff
@@ -156,6 +158,7 @@ export function useComponentWrapper(services: EditorServices): ComponentWrapperT
 
     const onClickWrapper = (e: React.MouseEvent<HTMLElement>) => {
       e.stopPropagation();
+      setRightTabIdx(0);
```
tanbowensg, line 161: Don't do setIndex here. There may be other place select component. Watch `selectedComponentId` change in `editorStore`.

packages/editor/src/services/EditorStore.ts
```diff
@@ -107,6 +111,14 @@ export class EditorStore {
     return this.currentComponentsVersion === this.lastSavedComponentsVersion;
   }
 
+  get leftTabIdx() {
+    return this._leftTabIdx;
```
tanbowensg, line 115: No need to duplicate. `hoverComponentId` and other two property is duplicated for special reasons.

packages/editor/src/services/enum.ts
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-unused-vars */
+enum ExploreMenuTabs {
```
tanbowensg, line 2: typo: explore -> explorer

tanbowensg (changes requested):

packages/editor/src/services/enum.ts
Yuyz0112: BTW, I think it's an eslint bug which shows enum as unused variables.

xzdry: > typo: explore -> explorer

corrected

xzdry: > BTW, I think it's an eslint bug which shows enum as unused variables.

yes, can use `"@typescript-eslint/no-unused-vars": ["error"]` instead of `no-unused-vars` to solve this problem

tanbowensg (approved):

tanbowensg:
ping @Yuyz0112

PR #244, opened 2022-01-29
improve state viewer UX

Yuyz0112:
1. watch state change and update the UI
2. use react-json-tree instead of CodeMirror


packages/editor/src/components/CodeEditor/StateViewer.tsx
```diff
@@ -0,0 +1,51 @@
...
+  base0C: '#a1efe4',
+  base0D: '#66d9ef',
+  base0E: '#ae81ff',
+  base0F: '#cc6633',
+};
+
+export const StateViewer: React.FC<{ store: Record<string, unknown> }> = ({ store }) => {
+  const [data, setData] = useState(store);
+  useEffect(() => {
+    watch(store, newValue => {
```
MrWindlike, line 32: May watch the store at `initSunmaoUIEditor` and pass it to all children is better? Because it may need the reactive store in otherwhere.

MrWindlike: For examples, in some components(Input) which has some states must declare two same states, one in the component and another one in the store. The codes look like that:
```js
const StateSchema = Type.Object({
  value: Type.String(),
});

...

const InputImpl = (props)=> {
  const [value, setValue] = useState(...);

  const onChange = (event: React.ChangeEvent<HTMLInputElement>) => { setValue(event.target.value); };

  useEffect(() => {
    mergeState({ value });
  }, [mergeState, value]);

  return ...;
}
```

That is too complex to manager the shared state. May make the store reactive in editor and just use store states in component is better.

tanbowensg: Maybe a little bit complex but not that complex. If not do so, `ComponentImpl` need get the `store` to update value. That gives `ComponentImpl` too much flexibility, which could make code hard to debug. Some Constraints make code easy to maintain.

Yuyz0112: Let me understand the question.

> Because it may need the reactive store in otherwhere.

> That is too complex to manager the shared state. May make the store reactive in editor and just use store states in component is better.

Do you mean let the input component remove its local state and change to a controlled component, which uses the global store as its state?

Actually, there are some design decisions in the current global state store.

1. We do not want to change the way users used to write a component. Because there should be many components were already implemented and just add a wrapper to be a sunmao-ui component.
2. We want the state transition in a sunmao-ui app to be explicit, e.g, we know which component changes the store and which component calls another component's method.

Under these premises, we implement the store as a **write-only** store for the components and a **read-only** store for the application.

In the view of the component developers, as the function name indicates, they are **merging component's** internal state to the global state on-demand. And in some cases, components' may even merge props to the global state.

Compared to using the global store as a shared state in a component(like redux or mobx):

1. A component only needs to make sure the states they declared in the schema will be merged to the global store. If we choose the shared state model, **components should make sure they will react when the store changes,** otherwise application builder may find changing the store directly only works for part of the component which is an unpredictable behavior.
2. A component explicitly declared how to interact with it remotely by defining the methods schema, which is an imperative interface instead of a declarative interface. The implementation is done by using the `subscribeMethods`.

And it's also easier to achieve type-safe and better static analysis by using the `mergeState` and `subscribeMethods`. The shortage is what you already point out: there is a learning curve to implement a sunmao-ui component.

In summary, this is a trade-off between the learning curve and building a predictable component interface, any further discussion and improvements should be very valuable.

@MrWindlike

MrWindlike: Got it. But should we make the store reactive in the React application and don't pass it to the lib components? Because we may use it to render in somewhere. If the store isn't reactive, may make some render mistakes in some situations.

Yuyz0112: Good point, so a `useStateStore` hook will be better than using `watch + store` everywhere?

MrWindlike: Yes, or may watching the store at the root component and making it as a reactive store then pass it down to all the children components be better? Because it can make sure the store I read from the `props` is reactive, which is more in accord with React's design.

Yuyz0112: updated

Yuyz0112: ping

MrWindlike: No problem.

tanbowensg (approved):
LGTM

Issue #225, opened 2022-01-25
feature: code auto-complete in the editor


Yuyz0112:
In sumao-ui we support writing an **expression** in the schema, which leverages JS and reactive for implementing the app business logic.

This feature tries to improve the UX of writing an expression in sunmao-ui's editor, by providing auto-complete to help users input code.

## Core

By investigating other similar products' implementation, we will choose [tern](https://ternjs.net/) to implement the core function of our auto-complete feature.

Tern is an extendable code-analysis engine for JavaScript, which integrates with code-mirror(the code editor we are using) quite well. And it supports defining custom definition so we can dynamically create a definition of our runtime store and users will get hints of state properties when they input.

An [example](https://github.com/appsmithorg/appsmith/blob/ba06d797de4e18fdb5a610335f275174d385d5c8/app/client/src/utils/autocomplete/customTreeTypeDefCreator.ts) of dynamically creating tern definition.

## UX

There are some trade-offs between providing a simple widget or a flexible widget to edit the schema.

For example, when we know a component's property has a type of boolean, we may provide a switch widget to input true or false as the property's value. This may make users happy in 95% of scenarios, but in the others, users may want to input an expression to create a boolean value from the runtime store, e.g, other components' state. If we provide the code editor widget instead, then the first kind of users may feel confused to input a boolean value by typing true/false.

So in the first version of auto-complete, we will add a flag widget to some of the schema widgets, which let users select whether to use a simple mode(like switch widget for boolean value) or an advance mode(code editor with auto-complete).

Examples:

simple mode:
![image](https://user-images.githubusercontent.com/13651389/150906365-76c6c970-bcaf-4ab0-9554-b686898e9082.png)

advance mode:
![image](https://user-images.githubusercontent.com/13651389/150906392-484f21af-5ec5-4e7e-985e-ea61599ffd67.png)

This flag is not recorded in the schema, so we need to get the default value of the flag when start editing a schema. The way of this is to check whether the value is a string value with expressions inside.

By the way, not all kinds of schema widgets need the flag.

- ✅ string field, enum need the flag, a general string will use the code editor by default
- ✅ object field
- ✅ array field
- ✅ boolean field
- ⛔ number field, use the code editor by default
- ⛔ null field, do not have a renderer
- ✅ multi schema field, use basic fields inside
- ⛔ unsupported field, no need to customize

tanbowensg:
In some cases, object and array need this flag too. For example, table data need to use the value of fetch trait.
[...]

PR #257, opened 2022-02-04
Auto complete: part 2

Yuyz0112:
#225 

1. Improve value serialization. Now we can use the expression editor in object and array values.
2. Fix multiline expression evaluation.


tanbowensg (approved):
LGTM

PR #276, opened 2022-02-14
fix: fix bugs with incorrect setting values and incorrect sorting

xzdry:
1. fix the bug that set the value incorrectly when the property name and title do not match
2. fix the bug with incorrect sorting based on weight

### DIFF
--- a/packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
+++ b/packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
@@ -32,14 +32,13 @@ const TopLevelField: React.FC<FieldProps> = props => {
     const properties = (schema.properties || {}) as Property;
     Object.keys(properties).forEach(name => {
       const schema = properties[name];
-      if (!schema.title) {
-        schema.name = name;
-      }
+      schema.name = name;
     });
     const grouped = groupBy(properties, c => c.category || 'Basic');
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
```diff
@@ -32,14 +32,13 @@ const TopLevelField: React.FC<FieldProps> = props => {
...
-        schema.name = name;
-      }
+      schema.name = name;
     });
     const grouped = groupBy(properties, c => c.category || 'Basic');
+
     return Object.keys(grouped).map(name => ({
       name,
-      schemas: sortBy(grouped[name], c => c.weight || -Infinity),
+      schemas: sortBy(grouped[name],'weight'),
```
Yuyz0112, line 41: what's the difference between the new way and the previous one?

xzdry: think 0 is more reasonable as the highest weighting

xzdry: I use smaller numbers as higher weights,and explicit fallback  is better,I've changed it to `c => c.weight ?? Infinity`

Yuyz0112 (approved):

PR #293, opened 2022-02-21
feat: add a preset order to property categories

xzdry:


packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
```diff
@@ -18,6 +18,23 @@ type ExpandSpec = {
   name?: string;
 };
 
+const PRESET_PROPERTY_CATEGORY_WEIGHT = {
+  Data: 100,
```
tanbowensg, line 22: These keys should be `enum` type instead of plain `string`.

packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
```diff
@@ -18,6 +18,23 @@ type ExpandSpec = {
...
+  Basic: 99,
+  Behavior: 3,
+  Layout: 2,
+  Style: 1,
+  Advance: -Infinity,
+};
+
+function getCategoryOrder(name: string): number {
+  return name in PRESET_PROPERTY_CATEGORY_WEIGHT
```
tanbowensg, line 31: Why not just `PRESET_PROPERTY_CATEGORY_WEIGHT[name] || 0`?

packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
xzdry: This is only used here to preset some order and will not be used elsewhere, so enum does not seem necessary

packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
xzdry: I copied this function from [ComponentList](https://github.com/webzard-io/sunmao-ui/blob/main/packages/editor/src/components/ComponentsList/ComponentList.tsx)，I think both expressions are fine

packages/editor/src/components/ComponentForm/JsonSchemaForm/TopLevelField.tsx
tanbowensg: It is very likely to be used elsewhere. It maybe used in chakra-lib and arco-lib. chakra-lib and arco-lib are using plain string to define category, which is not safe.

xzdry: We can export an enum in editor like
```
export enum PRESET_PROPERTY_CATEGORY{
  Data = 'Data',
  Basic = 'Basic',
  // ...
}
```
But it has no category weighting information to sort,a variable for marking weights is still needed here
or
```
export enum PRESET_PROPERTY_CATEGORY{
  Data = 100,
  Basic = 99,
  // ...
}
```
And access to category name via `PRESET_PROPERTY_CATEGORY[100]` when `groupby`

Or  other better way?

tanbowensg: I think the former way is better. We will have a enum of category name and a weight map. Only the enum need to be exported. Developed won't need weight map in component libs.

xzdry: Changed, later I will update chakra-lib and arco-lib in free time

tanbowensg (approved):

PR #297, opened 2022-02-21
improve the expression parser

Yuyz0112:
By introducing a looking ahead algorithm, now we can correctly parse expressions like `{{{a: 1}}}`.


tanbowensg (approved):
LGTM

PR #307, opened 2022-02-25
Make sure event handler is called in sync order and with the lateset state.

tanbowensg:
# Fixed Bugs
1. Fix: event handler is called async when `wait.time` is `0`.
2. Fix: `deepEvalAndWatch` method generates mutable object.

# Changes
In order to get the latest state, every time before an event handler is going to be called, **the handler schema need to be evaled again**. Otherwise, the event handler will be called with state of last frame.

This logic needs to be added to component **manually**. But only few components need to add it because these components need to trigger events by itself, like button in `table` or `onComplete` in `fetch`. Most components' events are handled by event trait.

### DIFF
--- a/packages/chakra-ui-lib/src/components/Table/Table.tsx
+++ b/packages/chakra-ui-lib/src/components/Table/Table.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { isArray, sortBy } from 'lodash-es';
 import {
   Table as BaseTable,
@@ -23,6 +23,7 @@ type SortRule = {
 
 export const TableImpl = implementTable(
   ({
+    component,
     data,
     majorKey,
     rowsPerPage,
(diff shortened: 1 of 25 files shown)


packages/chakra-ui-lib/src/components/Table/Table.tsx
```diff
@@ -184,13 +185,14 @@ export const TableImpl = implementTable(
...
                   {isMultiSelect ? checkbox : undefined}
-                  {columns.map(column => (
+                  {columns.map((column, j) => (
                     <TableTd
                       index={i}
                       key={column.key}
                       item={item}
-                      column={column}
                       onClickItem={() => selectItem(item)}
+                      rawColumn={(component.properties.columns as any)[j]}
```
Yuyz0112, line 194: why do we need `any` here?

packages/chakra-ui-lib/src/components/Table/TableTd.tsx
```diff
@@ -44,15 +52,18 @@ export const TableTd: React.FC<{
...
-            name: handler.method.name,
-            parameters: handler.method.parameters,
+            componentId: evaledHandler.componentId,
+            name: evaledHandler.method.name,
+            parameters: evaledHandler.method.parameters,
           });
         });
       };
-      content = <Button onClick={onClick}>{column.buttonConfig.text}</Button>;
+      content = <Button onClick={onClick}>{buttonConfig.text}</Button>;
```
Yuyz0112, line 66: Not related to this PR: do we still need a build-in button integration when we support module renderer? Especially when the module renderer can accept any component as a module.

tanbowensg: I think there could be these reasons:
1. Module is not stable now.
2. Button in table is very common and build-in button is much more convinient than a module.
3. Create a module only for a button maybe too heavy.

packages/chakra-ui-lib/src/components/Table/Table.tsx
tanbowensg: fixed.

Yuyz0112 (approved):

PR #311, opened 2022-02-28
feat: implement core space component

xzdry:


packages/runtime/src/components/_internal/Space.tsx
```diff
@@ -0,0 +1,101 @@
...
+    'flex-start',
+    'flex-end',
+    'center',
+    'space-between',
+    'space-around',
+  ],{
+    title:'Justify Content',
+    category:'Layout'
+  }),
+  size: Type.Union([StringUnion(['mini', 'small', 'middle', 'large']), Type.Number()],{
```
tanbowensg, line 25: I think just number type is enought? We don't have 'mini', 'small', 'middle', 'large' in core lib.

packages/runtime/src/components/_internal/Space.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { ReactNode } from 'react';
+import { css, CSSObject } from '@emotion/css';
+import { Static, Type } from '@sinclair/typebox';
+import { StringUnion } from '../../utils/stringUnion';
+
+export const SpacePropertySchema = Type.Object({
```
tanbowensg, line 6: The content of this file can be moved to `core/Space.tsx`. It is not internal.

tanbowensg (changes requested):
I think we should use the name `Stack` instead of `Space`. `Stack` is more proper than `Space`.

xzdry:
changed

packages/runtime/src/components/core/Stack.tsx
```diff
@@ -0,0 +1,111 @@
...
+      alignItems: align,
+      justifyContent: justify,
+      flexDirection: direction === 'vertical' ? 'column' : 'row',
+      display: 'inline-flex',
+      columnGap: horizontalSize,
+      rowGap: verticalSize,
+    };
+
+    if (wrap) {
+      style.flexWrap = 'wrap';
```
tanbowensg, line 60: Why not just adding this line the declaration of `style`?

packages/runtime/src/components/core/Stack.tsx
```diff
@@ -0,0 +1,111 @@
...
+export type StackProps = Static<typeof StackPropertySchema> & {
+  cssStyle?: string;
+};
+
+const Stack = React.forwardRef<HTMLDivElement, StackProps & { children: ReactNode }>(
+  (
+    { cssStyle, align, direction = 'horizontal', spacing = 12, children, wrap, justify },
+    ref
+  ) => {
+    const [horizontalSize, verticalSize] = React.useMemo(
```
tanbowensg, line 45: Is this unnecessary?

tanbowensg (changes requested):

packages/runtime/src/components/core/Stack.tsx
xzdry: There was other logic that has since been removed, so I'll change it

packages/runtime/src/components/core/Stack.tsx
xzdry: Do you mean useMemo or  support for array types of spacing?

tanbowensg: I mean spacing is `number` in spec, so it is unnecessary to detect if it is array.

xzdry: changed

tanbowensg (approved):

Issue #291, opened 2022-02-21
EditorMask can not work in dialog or other popover.


tanbowensg:
**What problem does this feature solve?**
Now mask cannot show in dialog or other popover.

PR #320, opened 2022-03-01
EditorMask support components in dialog

tanbowensg:
1. EditorMask support components in dialog #291 
2. fix: cannot drop component after components which have no slots

This implementation is kind of workaround for now. The key point is to filter the elements in dialog by detecting whether an element is child of `dialogContainerElement`.
The limitation of this way is that it cannot detect modals which append to body or multiple modals situation.
The ultimate solution maybe adding a z-axis in `rectMap`. But since we don't have multiple dialogs or popover for now, so we can implement it in the future.

### DIFF
--- a/packages/runtime/src/types/application.ts
+++ b/packages/runtime/src/types/application.ts
@@ -20,6 +20,7 @@ export type GridCallbacks = {
 
 export type ComponentParamsFromApp = {
   gridCallbacks?: GridCallbacks;
+  hooks?: AppHooks;
 };
 
 export type AppProps = {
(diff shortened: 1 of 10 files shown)


packages/runtime/src/types/application.ts
```diff
@@ -20,17 +20,21 @@ export type GridCallbacks = {
...
-  hooks?: AppHooks;
 } & ComponentParamsFromApp;
 
 export type AppHooks = {
+  // after app first render
   didMount?: () => void;
+  // app updates after schema changes
   didUpdate?: () => void;
+  // app updates after html elements change
+  didHtmlUpdate?: () => void;
```
Yuyz0112, line 39: maybe naming like this?

didUpdate -> willUpdate
didHtmlUpdate -> didUpdate

or
didHtmlUpdate -> didDomUpdate

tanbowensg: OK, I changed to `didDomUpdate`.

PR #322, opened 2022-03-02
feat: implement FormStack and FormControl

xzdry:
1.  Implement  `FormStack` to control the overall form layout
2. Implement `FormControl`    

demo:
image
image
image

### DIFF
--- a/packages/arco-lib/src/components/Form/DragComponentTips.tsx
+++ b/packages/arco-lib/src/components/Form/DragComponentTips.tsx
@@ -1,0 +1,15 @@
+import { Typography } from '@arco-design/web-react';
+
+const DragComponentTips: React.FC<{ componentName: string }> = ({ componentName }) => {
+  return (
+    <Typography.Paragraph style={{ color: '#b2b2b2' }}>
+      Please drag{' '}
+      <Typography.Text bold style={{ color: '#777' }}>
+        {componentName}
+      </Typography.Text>{' '}
+      components here
+    </Typography.Paragraph>
+  );
+};
+
+export { DragComponentTips };
--- a/packages/arco-lib/src/components/Form/FormControl.tsx
+++ b/packages/arco-lib/src/components/Form/FormControl.tsx
@@ -1,0 +1,95 @@
+import { Form } from '@arco-design/web-react';
+import { ComponentImpl, implementRuntimeComponent, Text } from '@sunmao-ui/runtime';
+import { css, cx } from '@emotion/css';
+import { Type, Static } from '@sinclair/typebox';
+import { FALLBACK_METADATA, getComponentProps } from '../../sunmao-helper';
+import { FormControlPropsSchema as BaseFormControlPropsSchema } from '../../generated/types/Form';
+import { DragComponentTips } from './DragComponentTips';
+import { FormControlErrorMessage } from './FormControlErrorMessage';
+
+const FormControlPropsSchema = Type.Object(BaseFormControlPropsSchema);
+const BaseFormControl = Form.Item;
+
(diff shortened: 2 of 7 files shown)


packages/editor/package.json
```diff
@@ -34,6 +34,7 @@
     "@chakra-ui/icons": "^1.0.15",
     "@chakra-ui/react": "^1.7.1",
     "@emotion/css": "^11.7.1",
+    "@sunmao-ui/arco-lib": "^0.0.7-alpha.1",
```
tanbowensg, line 37: don't depend on alpha version, use 0.0.6

packages/arco-lib/src/components/Form/FormControl.tsx
```diff
@@ -0,0 +1,95 @@
...
+    ),
+  };
+
+  return (
+    <BaseFormControl
+      label={<Text cssStyle="display:inline-block" value={label} />}
+      ref={elementRef}
+      {...formControlProps}
+      {...cProps}
+      labelAlign={cProps.labelAlign as any}
```
tanbowensg, line 48: Why as any?

packages/arco-lib/src/components/Form/DragComponentTips.tsx
```diff
@@ -0,0 +1,15 @@
+import { Typography } from '@arco-design/web-react';
+
+const DragComponentTips: React.FC<{ componentName: string }> = ({ componentName }) => {
```
tanbowensg, line 3: Can we change name to `EmptyPlaceholder`? I think it is more proper.

packages/arco-lib/src/generated/types/Form.ts
```diff
@@ -0,0 +1,73 @@
...
+        category: PRESET_PROPERTY_CATEGORY.Layout
+    })
+}
+
+export const FormPropsSchema = {
+    layout: StringUnion(['horizontal', 'vertical', 'inline'], {
+        title: 'Layout',
+        category: PRESET_PROPERTY_CATEGORY.Layout
+    }),
+    childrenLayout: StringUnion(['vertical', 'horizontal'], {
```
tanbowensg, line 57: What's the difference between `childrenLayout` and `layout`?

tanbowensg (changes requested):

packages/arco-lib/src/generated/types/Form.ts
xzdry: I wanted to have each element in the `FormControl` on a separate line, but setting each `FormControl` individually would have been cumbersome, so I provided this property to be set uniformly in the `Form Stack`

packages/arco-lib/src/components/Form/FormControl.tsx
xzdry: Both `Form Stack` and `Form Control` can set this property, but when Form Control has a default value('right' or 'left'), the one on Form Stack will not work, so you can't set labelAlign uniformly for all Formcontrol
So I added an 'unset' that arco didn't have as a default value for `labelAlign`

packages/arco-lib/src/generated/types/Form.ts
xzdry: I wanted to have each element in the `FormControl` on a separate line, but setting each `FormControl` individually would have been cumbersome, so I provided this property to be set uniformly in the `Form Stack`

tanbowensg: I think that will be confusing. We'd better let the `FormControl` itselft to control its layout.

tanbowensg (approved):

Issue #313, opened 2022-02-28
Proposal: Custom Widget


MrWindlike:
## Demand Overview
We would allow component developers to customize the form widgets for the properties. And we may support these features for the widget:

- Registering Widgets
- Widget Props
- Using Other Widgets In The Custom Widget
- Using One Widget To Configure Multiple Properties
- Widget Options

### Registering Widgets
We should register the widgets like registering the components, traits, and so on. Thus we also define a spec for widgets to describe their information:
```typescript
interface WidgetSpec {
  version: string;
  metadata: {
    name: string;
  }
}
```

### Widget Props
The interface of props which would pass to widget component: 
```typescript
interface WidgetProps {
  // the property schema
  schema: Schema & EditorSchema;
  // the value of the property
  value: any;
  // runtime services
  readonly services: Service;
  // components in the editor
  readonly components: Component[];
  // the handler of value changed
  onChange: (value: any)=> void;
  // render the sub-property schema normally
  renderBySchema: ({schema: Schema, onChange: (value: any)=> void, value: any})=> ReactElement;
}
```

### Using Other Widgets In The Custom Widget
If component developers want to use other widgets for sub-properties, they can define the widgets in the sub-properties spec options and use `renderBySchema` function in the custom widget component.

In The following, I will show you an example of implementing a custom widget for the `Table` component's `column` property.

First of all, we should define the `column` property's spec.
```javascript
{
  "column": Type.Array(Type.Object({
    ...,
    // use the common widget by its type
    "key": Type.KeyOf(...),
    // use the another widget defined in spec options
    "handlers": Type.Object({}, { "widget": "EventHandler" }),
  }, {
    "widget": "TableColumn"
  }))
}
```
And then we should implement the widget component and call `renderBySchema` inside.
```jsx
function TableColumn (props) {
  const { schema, value, onChange, renderBySchema } = props;
  
  const onValueChange = (key, value)=> {
    onChange({
      ...value,
      [key]: value
    })
  }
  
  return (
    <div>
      {/* implement custom widget here */ }
      {/* use others widgets for some properties */ }
      { renderBySchema({ 
          schema: schema.properties.key, 
          value: value.key,
          onChange: (v)=> onValueChange('key', value) 
      }) }
      { renderBySchema({
          schema: schema.properties.handlers,
          value: value.handlers,
          onChange: (v)=> onValueChange('handlers', value) 
      }) }
    </div>
  )
}
```

[...]

PR #325, opened 2022-03-02
feat: custom widget

MrWindlike:
Changes:
- [x] Adding new `@sunmao-ui/editor-sdk` package
- [x] Adding ChakraUI components into the kit object and exporting it
- [x] Changing all the form widgets according to the proposal
- [x] Fix the editor crash problem when the default values are not right 

Ref ISSUE: https://github.com/webzard-io/sunmao-ui/issues/313


packages/editor/src/AppModel/AppModel.ts
```diff
@@ -12,6 +13,7 @@ import {
 import { genComponent } from './utils';
 
 export class AppModel implements IAppModel {
+  widgets: Record<string, Widget> = {};
```
tanbowensg, line 16: `AppModel` is the model of application schema. Widgets are not part of application schema, so they should not be registered here.

MrWindlike: OK, I will use a new model to manage the widgets.

packages/editor-sdk/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "files": [
+    "dist",
+    "lib"
+  ],
+  "scripts": {
+    "build": "tsup src/index.ts --format cjs,esm,iife --legacy-output --no-splitting --clean --sourcemap --platform browser",
+    "typings": "tsc --emitDeclarationOnly",
+    "prepublish": "npm run build && npm run typings"
+  },
+  "dependencies": {
```
tanbowensg, line 21: add peer dependencies like other project.

packages/editor/package.json
```diff
@@ -34,8 +34,10 @@
     "@chakra-ui/icons": "^1.0.15",
     "@chakra-ui/react": "^1.7.1",
     "@emotion/css": "^11.7.1",
+    "@sinclair/typebox": "^0.21.2",
     "@sunmao-ui/chakra-ui-lib": "^0.2.0",
     "@sunmao-ui/core": "^0.4.0",
+    "@sunmao-ui/editor-sdk": "^0.5.0-alpha1",
```
tanbowensg, line 40: Did `editor-sdk` publish this version?

packages/editor/src/components/CodeEditor/CssEditor.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useEffect, useRef } from 'react';
 import CodeMirror from 'codemirror';
-import { Box } from '@chakra-ui/react';
+import { Box } from '@sunmao-ui/editor-sdk';
```
tanbowensg, line 3: I think we should not import `Box` and other chakra component from `editor-sdk`. Becase they are not widget, just common ui components.
And I don't think we will update these components in `editor-sdk`. 
Importing components from 2 packages is confusing.

packages/editor-sdk/src/types/widget.ts
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { JSONSchema7 } from 'json-schema';
+import { Component, ComponentSchema } from '@sunmao-ui/core';
+import { EditorServices } from './editor';
+
+export type Schema = Component<string, string, string, string>['spec']['properties'];
```
tanbowensg, line 6: It is JSONSchema7

packages/editor-sdk/src/types/widget.ts
```diff
@@ -0,0 +1,39 @@
...
+  widgetOptions?: WidgetOptions;
+  // category
+  category?: string;
+  weight?: number;
+  name?: string;
+};
+
+export type WidgetProps<WidgetOptions = Record<string, any>> = {
+  component: ComponentSchema;
+  schema: Schema & EditorSchema<WidgetOptions>;
```
tanbowensg, line 20: I guess here is actually spec. The spec of component properties.

packages/editor-sdk/src/types/widget.ts
```diff
@@ -0,0 +1,39 @@
...
+  services: EditorServices;
+  level: number;
+  value: any;
+  onChange: (v: any) => void;
+};
+
+export type WidgetOptions = {
```
tanbowensg, line 27: The name should be `Widget` according to the naming rule in `core`.

packages/editor-sdk/src/types/widget.ts
```diff
@@ -0,0 +1,39 @@
...
+  version: string;
+  metadata: {
+    name: string;
+  };
+  spec?: {
+    options?: JSONSchema7;
+  };
+};
+
+export type Widget<T = Record<string, any>> = WidgetOptions & {
```
tanbowensg, line 37: It should be `ImplementedWidget`

tanbowensg (changes requested):

packages/editor/src/components/CodeEditor/CssEditor.tsx
MrWindlike: But the `editor-sdk` should provide the same base editor UI components for developing widgets. Otherwise, the 
 widget developers need to use chakra UI in their projects.

Maybe change to `import { Box } from '@sunmao-ui/edtior-sdk/ui` could make it clear?

packages/editor/package.json
MrWindlike: Has not yet been, but the `lerna` would parse it as a local package. What version of `editor-sdk` should use here?

tanbowensg: 0.0.0 for now.

packages/editor/src/components/CodeEditor/CssEditor.tsx
tanbowensg: > Maybe change to import { Box } from '@sunmao-ui/edtior-sdk/ui could make it clear?
That will not help either. Since we don't have better idea, don't change it.

MrWindlike:
@tanbowensg OK. I think there is a problem with GitHub displaying the changes. After I solve the conflicts the changes become correct.

tanbowensg (approved):

PR #331, opened 2022-03-07
implement localstorage trait

xzdry:
1. implement localstorage trait
2. fix a bug that two updates to the store do not trigger a view refresh

### DIFF
--- a/packages/editor/src/components/CodeEditor/StateViewer.tsx
+++ b/packages/editor/src/components/CodeEditor/StateViewer.tsx
@@ -36,17 +36,17 @@ const style = css`
 
 export const StateViewer: React.FC<{ store: Record<string, unknown> }> = ({ store }) => {
   const [filterText, setFilterText] = useState('');
-  const [refreshFlag, setRefreshFlag] = useState(false);
+  const [refreshFlag, setRefreshFlag] = useState(0);
   const data = useMemo(() => {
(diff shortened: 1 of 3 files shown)


packages/editor/src/components/CodeEditor/StateViewer.tsx
```diff
@@ -36,17 +36,17 @@ const style = css`
...
       return filterText ? key.includes(filterText) : true;
     });
-  // eslint-disable-next-line react-hooks/exhaustive-deps
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [store, filterText, refreshFlag]);
 
   useEffect(() => {
     const stop = watch(store, () => {
-      setRefreshFlag(v => !v);
+      setRefreshFlag(v => v + 1);
```
tanbowensg, line 49: Why can this change fix the bug?

xzdry: We use a vue proxy to listen to the store, and if we `mergeState` again during an update, it will trigger two updates
I think react's `SyntheticEvent` will merge the two updates (setRefreshFlag twice and return the initial value) so that the view is not updated    
Here is an example(data is updated, but the view is not): https://stackblitz.com/edit/react-dmnplm?file=src/App.js

tanbowensg (approved):

PR #335, opened 2022-03-08
feat: add the popover widget

MrWindlike:
Changes:
- Adding a new field `path` to widget
- Adding a new popover widget
- Changing array widget

Screenshot:

image

### DIFF
--- a/packages/editor-sdk/src/components/Widgets/ArrayField.tsx
+++ b/packages/editor-sdk/src/components/Widgets/ArrayField.tsx
@@ -1,48 +1,17 @@
 import React from 'react';
 import { SchemaField } from './SchemaField';
-import { PopoverWidget } from './PopoverWidget';
 import { WidgetProps } from '../../types/widget';
 import { implementWidget, mergeWidgetOptionsIntoSchema } from '../../utils/widget';
 import {
-  Box,
-  ButtonGroup,
   IconButton,
-  Table,
-  Thead,
-  Tbody,
-  Tr,
-  Th,
-  Td,
+  Flex
 } from '@chakra-ui/react';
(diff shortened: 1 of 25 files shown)


packages/editor-sdk/src/components/Widgets/ArrayField.tsx
```diff
@@ -36,81 +68,146 @@ export const ArrayField: React.FC<WidgetProps<ArrayFieldWidgetOptionsType>> = pr
...
+      </ButtonGroup>
+    );
+
+    return {
+      props,
+      buttons,
+    };
+  });
+
+  return isNotBaseType ? (
```
tanbowensg, line 142: This return is too long and has too many ternary operator. Can we make this more readable?

tanbowensg: Or maybe we can split baseType array and non-baseType array to 2 components?

MrWindlike: Do you mean split it into two widgets? I think it is not good to do that because we should auto display the fittable content depending on different array types.

I have no good idea to simplify it now because the two type arrays need to reuse the buttons and sub props.

tanbowensg: I mean it can be one widget, while the implementation can be split to 2 or more components. This is for increacing the maintainability.

tanbowensg: Spliting code is not the only way to increase maintainability. Just removing the nested ternary operators also helps that.

MrWindlike: Updated.

PR #346, opened 2022-03-15
Fix/windlike patch

MrWindlike:
Changes: 
- make the widgets change value correctly
- make the API form's `disabled` can be expression
  image
- support expression for API form's body
  image
- add `setValue` method to localStorage trait


tanbowensg (approved):
LGTM

PR #347, opened 2022-03-16
feat: add the TableColumn widget

MrWindlike:


packages/chakra-ui-lib/package.json
```diff
@@ -23,7 +23,7 @@
   "scripts": {
     "dev": "vite",
     "test": "jest",
-    "build": "tsup src/index.ts --format cjs,esm,iife --legacy-output --inject ./react-import.js --clean --no-splitting --sourcemap",
+    "build": "tsup src/index.ts src/widgets/index.ts --format cjs,esm,iife --legacy-output --inject ./react-import.js --clean --no-splitting --sourcemap",
```
tanbowensg, line 26: why not import `src/widgets/index.ts` in `src/index.ts` ?

MrWindlike: The widgets are only used in the editor but not runtime. If export the widgets in the same entry file, it would make the size of the component lib entry file become bigger.

tanbowensg (approved):

PR #348, opened 2022-03-16
improve table component and editor area

xzdry:
1. add row click to table
2. component is clipped if necessary to fit the editor area

https://user-images.githubusercontent.com/96771399/158524393-d4880daa-492e-4371-b5a0-bee57e7d6f9e.mov


packages/arco-lib/src/components/Table.tsx
```diff
@@ -21,6 +27,7 @@ import {
 import { TableInstance } from '@arco-design/web-react/es/Table/table';
 
 const TableStateSchema = Type.Object({
+  clickedRow:Type.Optional(Type.Any()),
```
tanbowensg, line 30: I think we don't have to distinguish `clickedRow` and `selectedRow`. We just need to select the row after clicking it. Does arco support manually select a row?

tanbowensg (approved):

PR #351, opened 2022-03-18
Fix/windlike patch

MrWindlike:


packages/runtime/src/utils/resolveChildrenMap.ts
```diff
@@ -58,6 +64,6 @@ export function resolveChildrenMap(components: RuntimeComponentSchema[]): {
 
   return {
     childrenMap,
-    topLevelComponents,
+    topLevelComponents: [...topLevelComponents, ...fetchTraitComponents],
```
tanbowensg, line 67: Why put fetch trait components to the end?

tanbowensg: I think this is not a simple bug. It requires some changes of runtime. Can you create a new issue and provide a minimal schema to reproduce this bug?

MrWindlike: OK. I create a new issue to describe this problem: https://github.com/webzard-io/sunmao-ui/issues/356

tanbowensg (approved):

Issue #304, opened 2022-02-23
Proposal: Conditional Rendering For Editor Properties Form


MrWindlike:
## Demand Overview
In Some situations, the component lib developers want to make the component property form whether to render depends on another property value. 

Such as the `Table` Component, the `buttonConfig` property should only render when the column's `type` is `button`.

## Scheme Design
Maybe there are three ways to implement this demand:
- Use `Type.Union` to define the properties spec
- Add a new field to properties spec options to decide the property form whether should render
- Let component lib developers use custom widgets to render the properties form

### `Type.Union`
Currently, we already can use `Type.Union` to define property spec to make the form rendering depends on its value.
```js
{
  "type": Type.Union(
    [
      Type.String({
        unionValue: 'text'
      }),
      Type.Object({
        buttonConfig: Type.Object({
          text: Type.String({
            title: 'Text',
          }),
          handlers: Type.Array(EventHandlerSchema),
        }, {
          title: 'Button Config',
          unionValue: 'button'
        }),
      }),
    ]
  ) 
}
```
But this spec would break the type checking because the TypeScript type of `type` in the above example would be: 
```ts
type T = string | { buttonConfig: {...} };
```
So, Maybe this way doesn't work?

### New Field For Conditional Rendering
We can add a new field to spec options to describe when to render this property form. The type definition is:
```ts
interface Condition {
  // the key of the dependent property
  dependsOn: string;
  // rendering form when the dependent property value is equal to this value
  value: string;
}

type UnionConditions = Condition[];
type IntersectionConditions = Condition[];
```
And the example specs are just like these:
```js
// render when: type === 'text' || type === 'button'
{
  "text": Type.String({
    "unionConditions": [{
      "dependsOn": "type",
      "value": "text"
    }, {
      "dependsOn": "type",
      "value": "button"
    }]
  })
}
```
```js
// render when: type === 'text' && enable === true
{
  "text": Type.String({
    "intersectionConditions": [{
      "dependsOn": "type",
      "value": "text"
    }, {
      "dependsOn": "enable",
      "value": true
    }]
  })
}
```
But, how to design the `dependsOn` is a question, we should split it into different situations to talk about it.

#### The Related Properties Are Under The Same Level
When the related properties are under the same level, it would be simple to design the properties spec. The `dependsOn` is set to the key of another property which is under the same level just alright. Such as:
```js
{
  "columns": Type.Array(
  	Type.Object({
      ...,
      "type": Type.KeyOf(...),
[...]

PR #362, opened 2022-03-30
feat: implement conditional rendering

MrWindlike:
Follow the proposal https://github.com/webzard-io/sunmao-ui/issues/304 to implement condition rendering, but the implementation has some differences from the proposal:
- Remove the `UnionConditions` and `IntersectionConditions` types.
- Use the `conditions` field in component property spec options to control conditional rendering.
- Use the `key` instead of `dependsOn`
- Only support using the same level properties' values now. We maybe add a `path` to implement cross levels conditions in the future.

The conditional types are these:
```ts
type BaseType = string | number | boolean;
type Field = { key: string };

// compare
type EQ = { value: BaseType };
type Not = { not: BaseType };
type GT = { gt: number };
type LT = { lt: number };
type GTE = { gte: number };
type LTE = { lte: number };
export type ComparisonOperator = Field & (EQ | Not | GT | LT | GTE | LTE);

// logic
type And = { and: (Logic | ComparisonOperator)[] };
type Or = { or: (Logic | ComparisonOperator)[] };
export type Logic = And | Or;

export type Condition = (ComparisonOperator | Logic);
```

Preview: 

https://user-images.githubusercontent.com/25414733/160872421-63d8f15a-0ca8-4341-90fe-abcb7a1d63c8.mov


tanbowensg:
LGTM

tanbowensg (approved):

Yuyz0112 (approved):

PR #366, opened 2022-04-06
use re-resizable instead of react-resizable &  add ellipsis and change column width functions

xzdry:
1. editor: use `re-resizable` instead of `react-resizable` because it is simpler to use and does not cause re-rendering of child elements
2. arco table: add ellipsis and column widths can be changed via set property or drag


packages/arco-lib/src/components/Table.tsx
```diff
@@ -178,6 +217,24 @@ export const Table = implementRuntimeComponent({
     direction: undefined,
   });
   const [filterRule, setFilterRule] = useState();
+  const [columns, setColumns] = useState<ColumnProperty[]>([]);
+
+  const handleResize = (
```
tanbowensg, line 222: Can we wrap the resize logic to a component? `Table.tsx` is big and complex enough.

packages/arco-lib/src/components/Table.tsx
```diff
@@ -213,100 +270,113 @@ export const Table = implementRuntimeComponent({
...
-        case 'button':
-          const handleClick = () => {
-            const rawColumn = (component.properties.columns as ColumnProperty[])[i];
-            if (!rawColumn.btnCfg) return;
-            const evaledButtonConfig = services.stateManager.deepEval(
-              rawColumn.btnCfg,
-              true,
-              { [LIST_ITEM_EXP]: record }
+  useEffect(() => {
+    setColumns(
```
tanbowensg, line 274: Why make columns to state and put it in useEffect?

packages/arco-lib/src/generated/types/Table.ts
```diff
@@ -32,6 +32,13 @@ export const ColumnSchema = Type.Object({
     category: Category.Basic,
     description: 'The text you want to display instead of raw text.',
   }),
+  width: Type.Optional(Type.Number({
+    title: 'Width',
+  })),
+  ellipsis:Type.Optional(Type.Boolean({
```
tanbowensg, line 38: What is this property for?

packages/arco-lib/package.json
```diff
@@ -36,11 +36,13 @@
     "eslint-plugin-react-hooks": "^4.3.0",
     "lodash-es": "^4.17.21",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react-dom": "^17.0.2",
+    "react-resizable": "^3.0.4"
```
tanbowensg, line 40: Why use react-resizable in table and use re-resizeable in editor?

tanbowensg (changes requested):

packages/arco-lib/src/generated/types/Table.ts
xzdry: If the cell content exceeds the length, it is automatically omitted and displays `...`    
I'll add a description

tanbowensg: But the code seems like if not `ellipsis`, the text will be hidden?
```
colItem = <span title={column.ellipsis ? text : ''}>{text}</span>;
```

packages/arco-lib/src/components/Table.tsx
xzdry: Once the width has been changed by dragging, the width of the column needs to be changed by callback

packages/arco-lib/package.json
xzdry:
```js
// react-resizable
<Resizable handle={<span />}>
    <div id="container"></div>
</Resizable>

// render as 
<div id="container" class="react-resizable">
    <div class="handle"></div>
</div>
```

```js
// re-resizable
<Resizable>
    <div id="container"></div>
</Resizable>

// render as 
<div>
    <div id="container"></div>
</div>
```
react-resizable does not change the container, which is more suitable for scenarios like table where you need to return `th` directly. re-resizable can also be implemented, but you need to specify the return `th` manually and add the corresponding styles and child elements through the props returned by arco.      
react-resizable is more complex to use and will cause the child component to be re-rendered if it is not wrapped in memo. Whereas react-resizable simply adds a div wrap and changes its width, which is simple and does not cause the child to be re-rendered.
Also react-resizable has a mis-touch in the editor, haven't checked why yet

packages/arco-lib/src/generated/types/Table.ts
xzdry: Only the title will be hidden, it will be shown when ellipsis is on and hovering

packages/arco-lib/src/components/Table.tsx
xzdry: The table dragging logic is deeply coupled with the column property and is not easily separated

packages/arco-lib/src/components/Table.tsx
xzdry: oh,I was wrong, it's just so that when you change the width in inspect it takes effect immediately

tanbowensg (approved):

PR #369, opened 2022-04-07
Recover explorer and fix module's bugs

tanbowensg:
Fix some bugs that will cause module unstable.


packages/runtime/src/components/_internal/ImplWrapper.tsx
```diff
@@ -20,18 +20,24 @@ const _ImplWrapper = React.forwardRef<HTMLDivElement, ImplWrapperProps>((props,
...
+    if (!isInModule) {
+      eleMap.set(c.id, ele);
+    }
     hooks?.didDomUpdate && hooks?.didDomUpdate();
   };
 
   useEffect(() => {
-    if (eleRef.current) {
+    // If a component is in module, it should not have mask, so we needn't set it
+    if (eleRef.current && !isInModule) {
```
Yuyz0112, line 32: is it possible to infer whether a component is in a module from the application schema?

tanbowensg: No, because the component schema in module and application are the same.

Yuyz0112:
LGTM, please resolve the conflicts, and then we can merge.

tanbowensg:
Conlict Resolved.

Yuyz0112 (approved):

PR #375, opened 2022-04-13
feat: add refresh button

MrWindlike:
- Refresh the app

https://user-images.githubusercontent.com/25414733/163127119-0e54b27f-8e26-44cf-a69d-fe00f6d3a383.mov

- Recover the app after fixing the errors

https://user-images.githubusercontent.com/25414733/163127320-43536026-871e-415c-9690-28008d449c81.mov

### DIFF
--- a/packages/editor/src/components/Editor.tsx
+++ b/packages/editor/src/components/Editor.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState, useEffect } from 'react';
+import React, { useMemo, useState, useCallback, useEffect } from 'react';
 import { Application } from '@sunmao-ui/core';
 import {
   GridCallbacks,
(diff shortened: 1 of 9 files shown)


tanbowensg:
What is the difference between `unmountEffects` and `effects`?

MrWindlike:
> What is the difference between `unmountEffects` and `effects`?

`effects` would be executed in every update, but the `unmountEffects` only be expected when the component is unmounted.

The `HasInitializedMap` should be deleted when the component is unmounted instead of updating.

tanbowensg:
But `effect` is also designed to run after component unmount. I think they can combine to one.

MrWindlike:
> But effect is also designed to run after component unmount. I think they can combine to one.

Do you mean that the `effects` don't need to be executed in every update, only be executed when the component unmount?

Or let the trait developers control the `effects` when to be executed? If return the same reference functions, the `effects` would be executed in unmount. If return different references, the `effects` would be executed in every update.

tanbowensg:
Conflicts.

tanbowensg (approved):

Issue #373, opened 2022-04-11
The fetch trait doesn't execute when the dependent states first time changed


MrWindlike:
**Version**
0.5.6

**Reproduction JSON schema**
```json
{
  "version": "sunmao/v1",
  "kind": "Application",
  "metadata": {
    "name": "some App"
  },
  "spec": {
    "components": [
      {
        "id": "table6",
        "type": "arco/v1/table",
        "properties": {
          "columns": [
            {
              "title": "Name",
              "dataIndex": "name",
              "sorter": true,
              "sortDirections": [
                "ascend",
                "descend"
              ],
              "type": "text",
              "filter": true,
              "displayValue": ""
            },
            {
              "title": "Salary",
              "dataIndex": "salary",
              "sorter": true,
              "filter": false,
              "type": "text",
              "displayValue": "",
              "btnCfg": {
                "text": ""
              }
            },
            {
              "title": "Time",
              "dataIndex": "time",
              "sorter": true,
              "filter": false,
              "type": "text",
              "displayValue": ""
            },
            {
              "title": "Link",
              "dataIndex": "link",
              "type": "link",
              "filter": true,
              "sorter": false,
              "displayValue": ""
            },
            {
              "title": "CustomComponent",
              "dataIndex": "customComponent",
              "type": "module",
              "filter": false,
              "sorter": false,
              "module": {
                "id": "clistItemName-{{$listItem.id}}",
                "handlers": [],
                "properties": [],
                "type": "core/v1/text"
              },
              "displayValue": ""
            }
          ],
          "data": [
            {
              "key": "key 0",
              "name": "Kevin Sandra0",
              "link": "link-A",
              "salary": 863,
              "time": "2021-2-11T9:10:45.437Z"
            },
            {
              "key": "key 1",
              "name": "xzdry1",
              "link": "link-B",
              "salary": 276,
              "time": "2021-1-11T19:10:45.437Z"
            },
            {
              "key": "key 2",
              "name": "xzdry2",
              "link": "link-B",
              "salary": 656,
              "time": "2021-4-11T7:10:45.437Z"
            },
            {
              "key": "key 3",
              "name": "Kevin Sandra3",
              "link": "link-B",
              "salary": 625,
              "time": "2021-10-11T1:10:45.437Z"
            },
            {
              "key": "key 4",
              "name": "xzdry4",
              "link": "link-A",
              "salary": 10,
              "time": "2021-10-11T13:10:45.437Z"
            },
            {
[...]

Issue #356, opened 2022-03-24
The states depended on fetch trait are not correct


MrWindlike:
**Version**
0.5.2

**Steps to reproduce**
When the fetch trait depends on other components' states behind it, it would be executed first and can't take the right state values.

The reproduce JSON schema:
```json
{
  "version": "sunmao/v1",
  "kind": "Application",
  "metadata": {
    "name": "some App"
  },
  "spec": {
    "components": [
      {
        "id": "Fetch",
        "type": "core/v1/dummy",
        "properties": {},
        "traits": [
          {
            "type": "core/v1/fetch",
            "properties": {
              "url": "/?value={{localStorage.value}}",
              "method": "get",
              "lazy": false,
              "headers": {},
              "body": {},
              "bodyType": "json",
              "onComplete": [],
              "onError": []
            }
          }
        ]
      },
      {
        "id": "localStorage",
        "type": "core/v1/dummy",
        "properties": {},
        "traits": [
          {
            "type": "core/v1/localStorage",
            "properties": {
              "key": "value",
              "value": "value"
            }
          }
        ]
      }
    ]
  }
}
```

**What is expected?**
The states' values depending on fetch trait should be correct whatever the sort is. At the above demo, the URL when fetching should be `/?value=value`.

**What is actually happening?**
![image](https://user-images.githubusercontent.com/25414733/159874706-57a5b3ba-eef5-4ba6-ae1f-4eda22b84682.png)

tanbowensg:
The root of this bug is that during the initialization of the application, the earlier components use the state of the later components. We could solve this problem with `initStateAndMethod` to some extent. But when it comes to the situations like `defaultValue` of the `input` component, it is does not help, because the `defaultValue` of the input is not declared on the spec, but in the property, and the time to merge it on the store is during the component rendering.

So for now, one way I can think of is based on order. We can build a dependency tree of components and reorder component according it. But I haven't decide whether this sorting logic is in editor or in runtime.

However, this method cannot cover all possibilities after all, such as:

1. Expression circular dependency of two components (this may be the wrong usage)
2. The parent component uses the state of the child component

MrWindlike:
> Could we set the defaultValue in the component spec for the states and init the states' values in initStateAndMethod to solve it?

Oh, I'm wrong. The component's `defaultValue` is a component property and its value depends on the schema.

tanbowensg:
[...]

PR #376, opened 2022-04-13
Remove hasFetched in fetch trait and add disabled property

tanbowensg:
1. Remove hasFetched 
2. add disabled property

related issues: #356 #373

### DIFF
--- a/packages/editor/src/components/DataSource/ApiForm/Basic.tsx
+++ b/packages/editor/src/components/DataSource/ApiForm/Basic.tsx
@@ -6,6 +6,7 @@ import {
   FormLabel,
   Switch,
   IconButton,
+  Text,
 } from '@chakra-ui/react';
 import { AddIcon } from '@chakra-ui/icons';
 import { FormikHelpers, FormikHandlers, FormikState } from 'formik';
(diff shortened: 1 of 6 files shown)


packages/runtime/src/traits/core/Fetch.tsx
```diff
@@ -24,14 +23,14 @@ const FetchTraitFactory: TraitImplFactory<Static<typeof FetchTraitPropertiesSpec
...
       subscribeMethods,
       componentId,
+      disabled,
     }) => {
-      const hashId = `#${componentId}@${'fetch'}`;
-      const hasFetched = hasFetchedMap.get(hashId);
       const lazy = _lazy === undefined ? true : _lazy;
 
       const fetchData = () => {
+        console.log('disabled', disabled);
```
Yuyz0112, line 31: no console

packages/editor/src/components/DataSource/ApiForm/Basic.tsx
```diff
@@ -145,6 +147,32 @@ export const Basic: React.FC<Props> = props => {
...
+          Disabled
+        </FormLabel>
+        <SpecWidget
+          component={api}
+          spec={Type.Boolean({ widgetOptions: { isShowAsideExpressionButton: true } })}
+          value={formik.values.disabled}
+          path={[]}
+          level={1}
+          services={services}
+          onChange={val => {
```
MrWindlike, line 162: The API form component's render is heavy, should use `useCallback` and `useMemo` to cache the `spec`, `path`, and `onChange` props to avoid unnecessary render.

packages/runtime/src/traits/core/Fetch.tsx
tanbowensg: fixed

MrWindlike (approved):

PR #378, opened 2022-04-14
Feat/arco

xzdry:
1. keep mounted of tab panels' content when not active
2. feat(badge): use conditional rendering
3. refactor the alert component and add events to the alert and avatar components

### DIFF
--- a/packages/editor/src/components/Editor.tsx
+++ b/packages/editor/src/components/Editor.tsx
@@ -219,6 +219,7 @@ export const Editor: React.FC<Props> = observer(
                 display="flex"
                 flexDirection="column"
                 textAlign="left"
+                lazyBehavior='keepMounted'
                 isLazy
                 index={explorerMenuTab}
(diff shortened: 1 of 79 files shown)


packages/editor/src/components/Editor.tsx
```diff
@@ -219,6 +219,7 @@ export const Editor: React.FC<Props> = observer(
                 display="flex"
                 flexDirection="column"
                 textAlign="left"
+                lazyBehavior='keepMounted'
```
tanbowensg, line 222: Can we make this as property? I think it is something like `visually` of hidden trait. Sometimes we don't want other tab components mount.

tanbowensg (changes requested):

packages/editor/src/components/Editor.tsx
xzdry: changed

tanbowensg (approved):

PR #380, opened 2022-04-15
New LifeCycle of Component and Trait

tanbowensg:
# New LifeCycle of Component and Trait

Trait itself has no lifetime because it's just a function. Whenever the Property of the Trait changes, a Trait is executed, which drives the Component to update once.

To use the life cycle for Trait, it must use the life cycle of Component. Specifically, the Trait returns the hook function, just like the current effect. Hook functions can be executed by a component or by an ImplWrapper (for most traits, the differences are small, except for hidden traits).

| Component Life Cycle  | What Components do                                                                 | What Traits do                   |
| --------------------- | ---------------------------------------------------------------------------------- | -------------------------------- |
| ImplWrapper render    | Run trait.                                                                         | Run Trait function.              |
| Component Render      | Run component's code.                                                              |                                  |
| Component useEffect   | mergeState, onRef                                                                  |                                  |
| ImplWrapper DidMount  | register ref<br>Eval trait's and component's property<br>Listen expression change. |                                  |
| ComponentDidMount     |                                                                                    | mergeState，send requests....     |
| ComponentDidUpdate    |                                                                                    | mergeState....                   |
| ImplWrapperDidUnmount | Remove self from store.                                                            |                                  |
| ComponentDidUnmount   |                                                                                    | Clear side effects if there are. |

# Refactor of ImplWrapper
1. The `ImplWrapper` will be refactored as above described. Some codes will run in other timing, for example, the timing of listenning expression change. The behavior will be different with before, but it is totally compatible with previous schema code.
2. `ImplWrapper` will be split to multiple files bacause it it too large to be in a single file.


packages/runtime/src/components/_internal/ImplWrapper/ImplWrapperMain.tsx
```diff
@@ -0,0 +1,171 @@
...
+// This hook will only run unmount function when unmount, not every time when unmount function changes.
+const useDidUnmount = (fn: Function) => {
+  const fnRef = useRef(fn);
+  fnRef.current = fn;
+
+  useEffect(() => () => fnRef.current(), []);
+};
+
+// This hook will run every times when component update, except when first render.
+const useDidUpdate = (fn: Function) => {
```
MrWindlike, line 159: Maybe we should support the dependencies and cleanup in the update? Because the trait may execute some logic and clean up some side-effects only when the dependencies are updated instead of any properties being updated.

For example, I want to implement a trait to achieve this:
```js
useEffect(()=> {
  // when the source or callback is changed, create the new subscription.
  const subscription = subscribe(props.source, props.callback);

  return () => {
    // unsubscribe the old source
    subscription.unsubscribe();
  };
}, [props.source, props.callback]);

```

tanbowensg: OK. I have added return functions to  `componentDidUpdate` and `componentDidMount`

MrWindlike (approved):

Yuyz0112 (approved):

PR #383, opened 2022-04-25
Add UnmountImplWrapper

tanbowensg:
# Changes
* Add `hasUnmount` field in trait metadata annotation
* Add `UnmountImplWrapper`.  `UnmountImplWrapper` will check whether it has `hasUnmount` traits in component before. run into `ImplWrapperMain`.

### DIFF
--- a/packages/core/src/trait.ts
+++ b/packages/core/src/trait.ts
@@ -5,7 +5,7 @@ import { parseVersion, Version } from './version';
 
 // spec
 
-type TraitMetaData = Metadata<{ hasUnmount: boolean }>;
+type TraitMetaData = Metadata<{ beforeRender?: boolean }>;
 
 export type Trait = {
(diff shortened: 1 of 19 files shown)


packages/core/src/trait.ts
```diff
@@ -5,10 +5,12 @@ import { parseVersion, Version } from './version';
 
 // spec
 
+type TraitMetaData = Metadata<{ hasUnmount: boolean }>;
```
Yuyz0112, line 8: `hasUnmount` sounds ad-hoc for this scenario, is there some general word to describe this pattern(execute before render/skip render)?

tanbowensg: ok, I have changed to `beforeRender`.

Yuyz0112 (approved):

Issue #388, opened 2022-04-29
Proposal: Slot Props


Yuyz0112:
### Status Quo

The slot is one of the core features of sunmao-ui, which helps organize the hierarchy of components.

To implement a slot, a component needs to declare slot names as an array of strings in its spec and call `slotsElements[slotName]` to get the components in the slots at runtime.

To put a component into a slot, the application spec will use the slot trait and describe the desired target slot in the container property.

This design is fully static and causes some issues.

For example, a Tab component may have an arbitrary amount of tabs, and users may want to put components into different tabs' slots.

Currently, we will declare a `content` slot in the Tab component's spec, and put all the components into this slot. And at runtime, let the Tab component pick components by index, which is fragile and implicit.

### How does the Slot Props feature work

Slot props is a new proposal to provide the dynamic ability to the slot feature.

With slot props, components can pass typed data to a component in their slots.

In the application, a component with the slot trait can access the slot props. And the slot trait itself can define existence by using slot props' value.

#### Architechture Overview

![image](https://user-images.githubusercontent.com/13651389/165899743-01ebc3c8-1b58-4562-97a6-ee93486f6cce.png)

#### Who knows the slot props' type

The component's spec knows.

Component builders know what kind of data will be passed to the slot, so they can describe the type of the data in the component's spec statically. For the same example, a Tab component will pass data `tabIndex` with type `number` to the slot.

The change of the component spec looks like this:

```diff
spec: {
- slots: ['content']
+ slots: {
+   content: {
+     slotProps: Type.Object({
+       tabIndex: Type.Number(),
+     }),
+   },
+ },
}
```

#### Who passes the data to a slot

The component's runtime does.

An example of the runtime code looks like this:

```tsx
<TabPanels>
  {tabNames.map((_, idx) => {
    return (
      <TabPanel key={idx}>
        {slotsElements?.content ? (
          <slotsElements.content tabIndex={idx} />
        ) : (
          placeholder
        )}
      </TabPanel>
    );
  })}
</TabPanels>;
```

To pass the props to a slot element, we need the slot element to be a React component instead of a React node. Also in the example code, we shared an intuitive way to implement slot default content(check the `placeholder` part).

#### How to access the slot props in an application

When an application component was put into a slot by using the slot trait, itself and all its traits will be put into a slot scope.

All the members in the slot scope can access its slot props. Slot props will be stored under the namespace `$slot`.

An example of the access looks like this:

```diff
{
  "id": "btn",
[...]

PR #389, opened 2022-05-05
impl #388, support slot props

Yuyz0112:
Implementation in this PR follows the design described in #388.

Currently, this status is ready for review but not ready for merge because it only migrates two components to the new slot spec.

Also, this implementation introduces a breaking change to the components spec, so we need a major release if we are going to merge this.


Yuyz0112:
CI will not pass before we migrate all the components.

tanbowensg:
I think we can introduce this pr next tuesday.

tanbowensg:
We need allow user to edit `ifCondition` in editor.

Yuyz0112:
@tanbowensg @xzdry I've migrated all the components to the latest slots spec.

And there are several components in the arco-lib that need check because I think it needs slot props but I do not know the usage of its slots.

- Cascader
- Tabs
- Collapse

xzdry:
> @tanbowensg @xzdry I've migrated all the components to the latest slots spec.
> 
> And there are several components in the arco-lib that need check because I think it needs slot props but I do not know the usage of its slots.
> 
> * Cascader
> * Tabs
> * Collapse

The `tabs` are fine. `collapse` needs to be changed to the following:
```tsx
export const Collapse = implementRuntimeComponent(options)(props => {
  const { defaultActiveKey, options, ...cProps } = getComponentProps(props);
  const { elementRef, mergeState, slotsElements, customStyle, callbackMap } = props;

  const [activeKey, setActiveKey] = useState<string[]>(defaultActiveKey.map(String));

  const onChange = useCallback(
    (currentOperateKey: string, activeKey: string[]) => {
      setActiveKey(activeKey);
      mergeState({ activeKey });
      callbackMap?.onChange?.();
    },
    [callbackMap, mergeState]
  );

  return (
    <BaseCollapse
      ref={elementRef}
      className={css(customStyle?.content)}
      {...cProps}
      activeKey={activeKey}
      onChange={onChange}
    >
      {options.map((o, idx) => {
        const { key, ...props } = o;
        return (
          <BaseCollapse.Item key={key} name={String(key)} {...props}>
            {slotsElements?.content ? (
              <slotsElements.content activeKey={String(key)} />
            ) : null}
          </BaseCollapse.Item>
        );
      })}
    </BaseCollapse>
  );
});

```

As for the `Cascader` component itself there seems to be some problems, I'll check it later and change it

Yuyz0112:
@tanbowensg 

I've optimized the render process to avoid the new slot elements object causing re-mount.

Now, the process looks like this:

render an app
  render a component
    get slot elements map for this component: the map will only change when props other than `slotProps` change.
      create `<ImplWrapper />` for slot element: it will keep the same instance if the slot elements map is not re-new.
        when slot props change, it will cause the `<ImplWrapper />` re-render.

If I understand correctly, you are trying to make the `<ImplWrapper />` static and state manager watchers should only cause all the re-render.
I like this idea, but currently, slot props break this design. Although this does not cause extra re-render, I'd love to re-implement it in the next PR by creating a slot props store and making this `<ImplWrapper />` static again.

tanbowensg:
> Although this does not cause extra re-render
In my testing, I found this pr will cause component re-render after its siblings component changes in editor. I worry about that this may lead to poor performance in heavy app, like cluster-installer. We need to publish a beta version to see how will it affect cluster-installer.

Yuyz0112:
> In my testing, I found this pr will cause component re-render after its siblings component changes in the editor.

This is not the expected behavior. Could you share a minimal schema to reproduce? I'm going to fix this before merging.

tanbowensg:
If I change `input3`'s property, everything is ok. If I change `input2`'s property, both `input1` and `input2` will remount.

https://user-images.githubusercontent.com/12260952/168257862-8c0c66bf-6a45-4d76-8da4-02afb3bc49e9.mov


```
{
  "version": "sunmao/v1",
  "kind": "Application",
  "metadata": {
    "name": "some App"
  },
  "spec": {
    "components": [
      {
        "id": "box1",
        "type": "chakra_ui/v1/box",
        "properties": {
          "w": 40,
          "h": 40,
          "minW": "",
          "maxW": "",
          "minH": "",
          "maxH": "",
          "m": "",
          "p": "",
          "overflow": "",
          "bg": "",
          "bgColor": "",
          "opacity": "",
          "bgGradient": "",
          "bgClip": "",
          "bgImage": "",
          "bgSize": "",
          "bgPosition": "",
          "bgRepeat": "",
          "bgAttachment": "",
          "boxShadow": "",
          "border": "1px solid black",
          "borderRadius": ""
        },
        "traits": [
          {
            "type": "core/v1/style",
            "properties": {
              "styles": []
            }
          }
        ]
      },
      {
        "id": "input1",
        "type": "arco/v1/input",
        "properties": {
          "allowClear": false,
          "disabled": false,
          "readOnly": false,
          "defaultValue": "",
          "placeholder": "please input",
          "error": false,
          "size": "default"
        },
        "traits": [
          {
            "type": "core/v1/slot",
            "properties": {
              "container": {
                "id": "box1",
                "slot": "content"
              }
            }
          }
        ]
      },
      {
        "id": "input2",
        "type": "arco/v1/input",
        "properties": {
          "allowClear": false,
          "disabled": false,
          "readOnly": false,
          "defaultValue": "",
          "placeholder": "please input",
          "error": false,
          "size": "default"
        },
        "traits": [
          {
            "type": "core/v1/slot",
            "properties": {
              "container": {
                "id": "box1",
                "slot": "content"
              }
            }
          }
        ]
      },
      {
        "id": "input3",
        "type": "arco/v1/input",
        "properties": {
          "allowClear": false,
          "disabled": false,
          "readOnly": false,
          "defaultValue": "",
          "placeholder": "please input",
          "error": false,
          "size": "default"
        },
        "traits": []
      }
    ]
  }
}
```

Yuyz0112:
@tanbowensg 

The re-mount issue is caused by creating react element inline when we collect slots elements. So I change the slot elements interface from the React element to a plain function with the React nodes as return values.

Now the usage of slot elements looks like this:
```
            {slotsElements?.content ? (
              slotsElements.content({ activeKey, index })
            ) : (
              <EmptyPlaceholder componentName="" />
            )}
```
and the re-mount issue has been fixed.

Yuyz0112:
@xzdry I've updated the Collapse component's implementation.

Yuyz0112:
still has some design needs a double-check, mark as a draft at this moment

packages/core/src/slot.ts
```diff
@@ -0,0 +1,5 @@
+import { JSONSchema7 } from 'json-schema';
+
+export type SlotSchema = {
```
MrWindlike, line 3: May name it `SlotSpec`?

tanbowensg: ok

PR #398, opened 2022-05-12
fix some bugs

xzdry:
1. Add `setActiveTab` methods to tabs component.
2. Keep the scroll bar of each panel of explore menu tabs independent,previously, each panel under a tabs component shared the scroll bar of the parent element, so it could not maintain its original position when switching between tabs.
3. Fixed a bug where repeatedly clicking on the currently edited application (repeatedly clicking on an app or module) would cause the `stateManager's` store to be cleared


tanbowensg (approved):
LGTM

PR #401, opened 2022-05-15
feat: improve the editor performance

MrWindlike:
## Introduction

The editor is slow rendering when selecting the component in the huge application, so I improve the performance in these ways:
- Lazy loading.
  Render the `Popover`, `Accordion` and etc. components when they are opened instead of mounted.
- Extract the changeable props from the heavy component to the HOC.
  The `selectedComponentId` prop of `ComponentTree` component would be changed which would make the whole tree update when selecting the component, but the `isSelected` may not change. Thus, I create a HOC to read the `selectedComponentId` in the store and calculate the `isSelected`.

### Before
The rendering's time before optimization takes 258.6ms.
before

### After
The rendering's time after optimization takes 74.3ms.
after

### DIFF
--- a/packages/editor/src/components/Editor.tsx
+++ b/packages/editor/src/components/Editor.tsx
@@ -145,6 +145,12 @@ export const Editor: React.FC<Props> = observer(
       }
     }, [isDisplayApp]);
     const onPreview = useCallback(() => setPreview(true), []);
+    const onSelectComponent = useCallback(id => {
+      editorStore.setSelectedComponentId(id);
+    }, [editorStore]);
+    const onRightTabChange = useCallback(activatedTab => {
+      setToolMenuTab(activatedTab);
+    }, []);
 
     const renderMain = () => {
       const appBox = (
@@ -216,9 +222,7 @@ export const Editor: React.FC<Props> = observer(
                     <StructureTree
                       components={components}
                       selectedComponentId={selectedComponentId}
-                      onSelectComponent={id => {
-                        editorStore.setSelectedComponentId(id);
-                      }}
+                      onSelectComponent={onSelectComponent}
                       services={services}
                     />
                   </TabPanel>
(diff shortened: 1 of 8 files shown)


packages/editor/src/components/Editor.tsx
```diff
@@ -257,9 +261,8 @@ export const Editor: React.FC<Props> = observer(
                   display="flex"
                   flexDirection="column"
                   index={toolMenuTab}
-                  onChange={activatedTab => {
-                    setToolMenuTab(activatedTab);
-                  }}
+                  onChange={onRightTabChange}
+                  isLazy
```
tanbowensg, line 265: It can not be lazy, because there was an feature that everytime when user change tab, the scroll position should remain the last position. So the tab contents must be mounted all the time. Right? @xzdry

tanbowensg:
Sunmao editor has a design principle that the whole editor can be refactor by sunmao itself which requires every part of editor should not be coupled. However, this hoc will break this principle because it directly import the `editorStore` instead of receiving it from props. 
But I have another way to achieve this. The hoc can be refactored to a component and it still receive `selectedComponentId` from props. In this component, the logic is the same with hoc. It computes the `isSelected` and pass it to `ComponentTree`. In this way, we don't have to import `editorStore` and still can memo the render result.

packages/editor/src/components/Editor.tsx
xzdry: Setting `isLazy` means that the tab panel's content will only be loaded when they are first activated, not all of them are mounted at the beginning
So there is no problem here, and I can subsequently control whether it keeps mounted when switching by setting `lazyBehavior`

packages/editor/src/components/StructureTree/ComponentTree.tsx
```diff
@@ -140,20 +161,26 @@ export const ComponentTree: React.FC<Props> = props => {
...
         />
       </DropComponentWrapper>
       {isExpanded ? slotsEle : undefined}
     </VStack>
   );
 };
+
+export const ComponentTreeWrapper: React.FC<Props> = observeSelected(
+  React.memo(ComponentTree, (prevProps, nextProps) => {
+    return (Object.keys(prevProps) as (keyof typeof prevProps)[]).every(
```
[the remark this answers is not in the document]
MrWindlike: Using the `React.PureComponent` needs to change the component to be the class component, I think we don't have to do that. I found the `React.memo` has the default shallow comparison function, so I remove the custom compare function.

tanbowensg (approved):
LGTM

PR #402, opened 2022-05-16
Feat/arco

xzdry:
1. feat(slider): add slider component
2. refactor(formcontrol): In a previous commit I removed the grid layout property from formcontrol, which was undesirable as it was still needed in most cases, so it was restored
3. feat(link): add a default target style to the link
4. feat(input,textarea): Add the updateWhenDefaultValueChanges property to change the default value accordingly

### DIFF
--- a/packages/arco-lib/src/components/Link.tsx
+++ b/packages/arco-lib/src/components/Link.tsx
@@ -11,12 +11,20 @@ const LinkStateSpec = Type.Object({});
 const exampleProperties: Static<typeof LinkPropsSpec> = {
   disabled: false,
   hoverable: true,
-  status: 'warning',
+  status: 'default',
   href: 'https://www.smartx.com/',
   content: 'Link',
 };
 
-const options = {
+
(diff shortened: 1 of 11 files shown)


packages/arco-lib/src/components/Link.tsx
```diff
@@ -11,12 +11,20 @@ const LinkStateSpec = Type.Object({});
...
 };
 
-const options = {
+
+const statusMap = {
+  default: undefined,
+  success: 'success',
+  error: 'error',
+  warning: 'warning',
+} as const;
```
tanbowensg, line 25: why as const?

xzdry: This is because Link's status property accepts the types `"success" | "error" | "warning" | undefined` Therefore the `as const ` keyword is needed to avoid literal types being widened(e.g. no going from "success" to string)

tanbowensg (approved):
LGTM

Issue #248, opened 2022-01-30
Add spec of UtilMethod


tanbowensg:
**What problem does this feature solve?**

Since UtilMethod is also a member of lib, it should have its spec now.

PR #422, opened 2022-05-30
feat(utilMethod): add spec to utilMethod

MrWindlike:
BREAKING CHANGE:
use `type` to replace `name` when calling the method

ISSUES CLOSED: #248

### DIFF
--- a/packages/core/src/utilMethod.ts
+++ b/packages/core/src/utilMethod.ts
@@ -1,0 +1,28 @@
+import { JSONSchema7 } from 'json-schema';
+import { Metadata } from './metadata';
+import { parseVersion, type Version } from './version';
+
+export type UtilMethodSpec = {
+  parameters: JSONSchema7;
+};
+
+export type UtilMethod = {
+  version: string;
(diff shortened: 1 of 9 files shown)


packages/core/src/utilMethod.ts
```diff
@@ -0,0 +1,28 @@
...
+import { Metadata } from './metadata';
+import { parseVersion, type Version } from './version';
+
+export type UtilMethodSpec = {
+  parameters: JSONSchema7;
+};
+
+export type UtilMethod = {
+  version: string;
+  king: 'UtilMethod';
```
tanbowensg, line 11: king -> kind

tanbowensg (approved):

tanbowensg (changes requested):
Why do we need to use `type` to replace `name` when calling the method?

MrWindlike:
> Why do we need to use type to replace name when calling the method?

The different libs may have the same names of util methods.

tanbowensg (approved):

Issue #140, opened 2021-11-26
Table supports render component directly


tanbowensg:
Now table can only render module. It is necessary to be able to render component directyl.

tanbowensg:
Not only table, every component that allows rendering module should also allow rendering component.

tanbowensg:
closed by: https://github.com/webzard-io/sunmao-ui/pull/431

PR #431, opened 2022-06-02
List and table support render component directly

tanbowensg:
### Changes
1. Add `core/v1/list`. This list supports render component directly. `chakra_ui/v1/list` remains untouched for compatibility reason. Considering that `chakra_ui/v1/list` does not have any special abilities, it is recommanded to use `core/v1/list` in the future.
2. Update `chakra_ui/v1/table` & `arco/v1/table`. Add `component` type as a new column type. Other column types are preserved for compatibility reason, although they can be replaced by component.

### How to Use
To render component in list or table, you should create component in list's or table's slot. The children will not render normally, instead, they will be rendered by list or table component. List will use childrent as template, duplicates many components dynamically and inject scope object by `slotProps` into them.

### Example
```
        {
          "id": "list",
          "type": "core/v1/list",
          "properties": {
            "listData": "{{ root.listData }}"
          },
          "traits": []
        },
        {
          "id": "listSlot",
          "type": "chakra_ui/v1/button",
          "properties": {
            "text": {
              "raw": "{{$slot.$listItem.name}}",
              "format": "md"
            }
          },
          "traits": [
            {
              "type": "core/v1/slot",
              "properties": {
                "container": {
                  "id": "list",
                  "slot": "content"
                }
              }
            }
          ]
        }
```

截屏2022-06-02 下午5 30 50
related issues: #140


packages/arco-lib/src/generated/types/Table.ts
```diff
@@ -198,6 +208,16 @@ export const ColumnSpec = Type.Object({
     title: 'Display Value',
     category: Category.Basic,
     description: 'The text you want to display instead of raw text',
+    conditions: [
```
MrWindlike, line 211: It should be changed to:
```js
    conditions: [
      {
        or: [
          {
            key: 'type',
            value: 'link',
          },
          {
            key: 'type',
            value: 'text',
          },
        ],
      },
    ],
```

tanbowensg: ok

MrWindlike (approved):

PR #437, opened 2022-06-08
feat(StyleTraitForm): add SpaceWidget

xzdry:
Implementation of SpaceWidget:

https://user-images.githubusercontent.com/96771399/172525190-518753ee-31a8-42e3-9f01-7a94718aa1c2.mov

There are still some things that need to be done:
- **IMPORTANT**: Optimize the performance when mousemove, currently there are some serious frame drops, I feel that it is caused by too many re-rendering of react components, and need to find ways to optimize it in the future
- The SpaceWidget panel needs to be responsive

### DIFF
--- a/packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget.tsx
+++ b/packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget.tsx
@@ -1,0 +1,55 @@
+import React from 'react';
+import { CORE_VERSION, StyleWidgetName } from '@sunmao-ui/shared';
+import { WidgetProps } from '../../../types/widget';
+import { implementWidget, mergeWidgetOptionsIntoSpec } from '../../../utils/widget';
+import { ExpressionWidget } from '../ExpressionWidget';
+import { Grid, GridItem } from '@chakra-ui/react';
+
+const SPACE_DIRECTIONS = ['top', 'right', 'bottom', 'left'];
+
+export const SpaceWidget: React.FC<WidgetProps<{}>> = props => {
+  const { value, onChange } = props;
+
+  return (
+    <div>
+      <Grid
+        templateRows="repeat(3, 1fr)"
+        templateColumns="repeat(3, 1fr)"
+        templateAreas={`". top ."
+                        "left . right"
+                        ". bottom ."`}
+        gap="1"
+      >
+        {SPACE_DIRECTIONS.map((direction, idx) => {
+          return (
+            <GridItem
+              maxW="120px"
+              minW="0px"
+              key={direction}
+              area={direction}
+              gridArea={direction}
+            >
+              <ExpressionWidget
+                {...props}
+                spec={mergeWidgetOptionsIntoSpec(props.spec, {
+                  compactOptions: { height: '32px' },
+                })}
+                value={value[idx]}
+                onChange={v => {
+                  onChange(v, direction);
+                }}
+              />
+            </GridItem>
+          );
+        })}
+      </Grid>
+    </div>
+  );
+};
+
+export default implementWidget({
+  version: CORE_VERSION,
+  metadata: {
+    name: StyleWidgetName.Space,
+  },
+})(SpaceWidget);
--- a/packages/editor-sdk/src/components/Widgets/index.ts
+++ b/packages/editor-sdk/src/components/Widgets/index.ts
@@ -17,6 +17,7 @@ import popoverWidgetSpec from './PopoverWidget';
 import sizeWidgetSpec from './StyleWidgets/SizeWidget';
 import fontWidgetSpec from './StyleWidgets/FontWidget';
 import colorWidgetSpec from './StyleWidgets/ColorWidget';
+import spaceWidgetSpec from './StyleWidgets/SpaceWidget';
 
 export * from './SpecWidget';
 export * from './ArrayField';
@@ -36,8 +37,10 @@ export * from './PopoverWidget';
 export * from './StyleWidgets/SizeWidget';
 export * from './StyleWidgets/FontWidget';
 export * from './StyleWidgets/ColorWidget';
+export * from './StyleWidgets/SpaceWidget';
 
 export const widgets: ImplementedWidget<any>[] = [
+  spaceWidgetSpec,
   specWidgetSpec,
   arrayFieldSpec,
   booleanFieldSpec,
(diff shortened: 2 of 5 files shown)


Yuyz0112:
wow

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
```diff
@@ -0,0 +1,127 @@
...
+          shadow="base"
+          zIndex="popover"
+          overflowY="auto"
+          width="300px"
+          _focus={{ boxShadow: 'none' }}
+        >
+          <VStack w="full">
+            <HStack w="full">
+              <InputGroup>
+                <Input
```
tanbowensg, line 75: We should use `ExpressionEditor` here.

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
```diff
@@ -0,0 +1,127 @@
...
+                    if (inputValue === 'Auto' && hasAuto) {
+                      onChange(inputValue);
+                    } else if (!isNaN(Number(inputValue))) {
+                      onChange(Number(inputValue));
+                    } else {
+                      setInputValue(value);
+                    }
+                  }}
+                />
+                <InputRightAddon>px</InputRightAddon>
```
tanbowensg, line 90: Does the unit have to be `px`? Could we allow user use other units?

tanbowensg (changes requested):

tanbowensg:
The drag interaction is very fancy. However, I think it might not be very useful because it is hard to set an accurate value by dragging. For example, if I want to set `margin: 16px`, it will be much quicker to enter `16px` for four times than dragging four times.

tanbowensg:
The preset values also need change. In most time we will not set margin or padding bigger than 40px. 4px, 8px, 12px, 16px are more frequently used.

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
MrWindlike: Using the `ExpressionWidget` instead, because it could display the expression value and error.

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceWidget.tsx
```diff
@@ -0,0 +1,201 @@
...
+    <>
+      <Grid {...containerStyle}>
+        <Grid {...marginStyle}>
+          <SpacePanel
+            text="MARGIN"
+            type="margin"
+            width="224px"
+            height="120px"
+            values={marginValues}
+            onChange={(v: SpaceValue[]) => {
```
MrWindlike, line 170: It should use the `useCallback` to generate the handler.

xzdry:
> The drag interaction is very fancy. However, I think it might not be very useful because it is hard to set an accurate value by dragging. For example, if I want to set `margin: 16px`, it will be much quicker to enter `16px` for four times than dragging four times.

I think it might be useful in some cases, e.g.
Not knowing exactly position where you want to put the element at first, and determining a general position by dragging

> because it is hard to set an accurate value by dragging       

This is the main problem at the moment, the dragging process cannot be increased exactly 1 px by 1 px. If it could be done more precisely and smoothly, I think drag  would be more advantageous for smaller space values like `16px` etc. So I will try to optimise it

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
xzdry: Of course, I will add support for common units later

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
xzdry: There are two considerations for not using `ExpressionWidget` here.
1. there seem to be very few scenarios where margin or padding is needed in association with other elements, so the implementation allows only the specified values to be entered. Illegal input, such as 'something wrong', is silently corrected to the previous value
2. the original intention of the style widget was to simply set some common css properties, so there may not be much need for expressions? And we can still use complex expressions via the css editor.

tanbowensg:
> Not knowing exactly position where you want to put the element at first, and determining a general position by dragging
I think use arrow key to adjust value 1px by 1px is more proper way to do it. We rarely adjust margin value in a large range. In most times we just want to see the difference between 1 or 2 px.

What do you think about it? @MrWindlike @Yuyz0112

xzdry:
> I think use arrow key to adjust value 1px by 1px is more proper way to do it.

This is also a good way to go and I will try it out

xzdry:
I've tried a few ways to optimise the dragging fluency problem, but none of them work well, except for directly modifying the dom element. And the difference between the two is obvious. Here is a demo

https://user-images.githubusercontent.com/96771399/174020571-04a815ac-1746-48dc-a89c-5a81582243fd.mov

Using the arrow key also causes the same problem
But modifying the dom directly doesn't seem to be a good solution, it doesn't change through sunmao and the editor mask doesn't update with it, is there any other way?
For these reasons, I have removed the ability to make changes in real time for now.
I also removed the shortcuts because I felt that the previous style was not good and I still need to think about this part.
Here is a current demo

https://user-images.githubusercontent.com/96771399/174022311-97386ae6-8c6c-4470-a9ac-eb43e0299317.mov

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
xzdry: done

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceWidget.tsx
xzdry: done

packages/editor-sdk/src/components/Widgets/StyleWidgets/SpaceWidget/SpaceEditPanel.tsx
tanbowensg: The `ExpressionWidget` is still necessary because it is a widget which means so it could be used in any other where, not just style trait form and there is no css editor in other place.

tanbowensg (approved):
LGTM

PR #438, opened 2022-06-09
fix some arco bugs

xzdry:
Remove useless attributes
Fix slot-related bugs
Fix popover wrong display of controlled mode


packages/arco-lib/src/components/Tabs.tsx
```diff
@@ -119,15 +119,21 @@ export const Tabs = implementRuntimeComponent({
...
-          {slotsElements?.content
-            ? slotsElements.content({
-                tabIndex: idx,
-              })
-            : null}
-        </TabPane>
-      ))}
+      {tabs.map((tabItem, idx) =>
+        tabItem.hidden ? null : (
+          <TabPane
```
tanbowensg, line 124: Should it be `TabPanel`?

xzdry: I'm not sure, Arco used `pane`, a word that means a piece, section or side of something, which also seems correct

Issue #436, opened 2022-06-07
Add tests to runtime


tanbowensg:
`ImplWrapper` is the most complex and important part of sunmao. However, it does not have test yet which means we will not know the potential effects of every change we make in `ImplWrapper`. That's why we need add tests for it.

PR #444, opened 2022-06-10
Test/runtime

tanbowensg:
### Changes
1. Add some tests for `ImplWrapper`, `List` and `ModuleRenderer` and does not modify their current logic.
2. Add  a `TestLib` which consists of some simple components and trait just for test.

### Testing principles
There are 2 principles when testing sunmao:
1. Test the components' rendering result rathen than their implementation details.
2. Test the component as whole rather than apart.

For example, `ImplWrapper` is a very complex component which has a lot of state, effects. It is hard to test itself. So I add a `TestLib` to test it and only focus on the rendering result of `ImplWrapper`. If the result is right, than I assume the whole `ImplWrapper` is working as expected.

issue: #436


packages/runtime/__tests__/ImplWrapper/ImplWrapper.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+    new Promise<any>(res => {
+      setTimeout(() => res(true), 30);
+    });
+
+  it('it will cause extra render', async () => {
+    const { App, stateManager } = initSunmaoUI({ libs: [TestLib] });
+    stateManager.noConsoleError = true;
+    const { unmount } = render(<App options={AsyncMergeStateSchema} />);
+    await waitFor(timeoutPromise);
+    expect(await screen.findByTestId('tester')).toHaveTextContent('4');
```
MrWindlike, line 111: May add a comment to explain to others why it would render four times?

packages/runtime/__tests__/ImplWrapper/ImplWrapper.spec.tsx
```diff
@@ -0,0 +1,132 @@
...
+    const newMergeStateSchema = produce(AsyncMergeStateSchema, draft => {
+      const temp = draft.spec.components[0];
+      draft.spec.components[0] = draft.spec.components[1];
+      draft.spec.components[1] = temp;
+    });
+    const { App, stateManager } = initSunmaoUI({ libs: [TestLib] });
+    stateManager.noConsoleError = true;
+    const { unmount } = render(<App options={newMergeStateSchema} />);
+    await waitFor(timeoutPromise);
+    expect(await screen.findByTestId('tester')).toHaveTextContent('5');
```
MrWindlike, line 127: Same as above.

tanbowensg:
Updated.

MrWindlike (approved):

PR #448, opened 2022-06-13
feat: change the reference properties when changing the component id

MrWindlike:
Updates:
1. Adding the `emitter` into the `AppModel`.
2. Using the `AppModel`'s `emitter` to notify the `FieldModel` to change its property when changing the component id.
3. Adding the AST nodes into `FieldModel`'s `refs` to generate the new expression.
4. Adding the `isComponentId` into spec options to declare a property is whether refer to a component id.
5. Adding the related tests.

ISSUE: https://github.com/webzard-io/sunmao-ui/issues/106


packages/editor/src/AppModel/FieldModel.ts
```diff
@@ -109,28 +137,85 @@ export class FieldModel implements IFieldModel {
...
+          delete this.refs[oldId];
+
+          return [escodegen.generate(node)];
+        }
+
+        return exp;
+      });
+      const value = expChunkToString(newExps);
+
+      this.componentModel._isDirty = true;
```
tanbowensg, line 217: Is is missing to set TraitModel isDirty here?

packages/editor/src/AppModel/FieldModel.ts
```diff
@@ -1,18 +1,38 @@
...
+import escodegen from 'escodegen';
+import { JSONSchema7 } from 'json-schema';
+
+type Flatten<Type> = Type extends Array<infer Item> ? Item : Type;
 
 export class FieldModel implements IFieldModel {
   isDynamic = false;
-  refs: Record<ComponentId | ModuleId, string[]> = {};
+  refs: Record<ComponentId | ModuleId, RefInfo> = {};
+  private nodes: Record<string, acorn.Node> = {};
```
tanbowensg, line 24: I think we could use more specific names. `refs` and `nodes` are too abstract here.

packages/editor/__tests__/model/fieldModel.spec.ts
```diff
@@ -70,4 +78,43 @@ describe('Field test', () => {
     field.update({ value: 'text' });
     expect(field.rawValue).toEqual({ data: { a: 2 }, value: 'text' });
   });
+
+  it('change component id', () => {
```
tanbowensg, line 82: Add more test cases to make sure that normal text will not be changed, like `'Please input here'`, `'{{ {input: 123} }}'`, `'{{ 'input' }}'` ,  `['input']`

packages/editor/src/AppModel/IAppModel.ts
```diff
@@ -104,6 +113,7 @@ export interface IComponentModel {
 
 export interface ITraitModel {
   // trait id only exists in model, doesn't exist in schema
+  appModel: IAppModel;
```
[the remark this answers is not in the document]
MrWindlike: I'm not sure should we expose the `appModel`, because I just follow the `ComponentModel` to expose it. What do you think?

packages/editor/src/AppModel/FieldModel.ts
MrWindlike: How about the `componentRefs` and the `astNodes`?

packages/editor/src/AppModel/IAppModel.ts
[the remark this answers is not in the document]
tanbowensg: The interface should be as simple as posible. It seems like the appModel is only used inside `TraitModel`, so I think it should be private. And so does `traitModel` and `componentModel` in `IFieldModel`.

packages/editor/src/AppModel/FieldModel.ts
tanbowensg: `astNodes` is good. `componentRefs` is still confusing. I think the problem is that `ref` is ambigious here.`RefInfo` have not only has `nodes`(seems like a combination of `astNode` and other), but also has `properties`(not sure what is it).

packages/editor/src/AppModel/FieldModel.ts
MrWindlike: Done.

packages/editor/__tests__/model/fieldModel.spec.ts
MrWindlike: Done.

packages/editor/src/AppModel/FieldModel.ts
```diff
@@ -109,28 +137,90 @@ export class FieldModel implements IFieldModel {
...
+              this.refComponentInfos[lastIdentifier]?.refProperties.push(path);
               break;
             default:
           }
         },
       });
     });
   }
+
+  private onReferenceIdChange({ oldId, newId }: AppModelEventType['idChange']) {
```
tanbowensg, line 186: This is the core part of the pr and it seems a little bit hard to understand. We should add some comments here to help us understand in the future.
Other things LGTM.

tanbowensg (changes requested):

tanbowensg (approved):
LGTM

PR #450, opened 2022-06-13
fix: trait form is covered by section titles

Mark-Fenng:
### Bug

If I click the button 'Add Trait', the popup form will be covered by some section titles with higher z index values.
image

### Fix

Change the z index value of that form to 100. '100' is equal to the z-index of section titles.
image

Effect applied this fix:
image


tanbowensg (approved):
Good!

PR #452, opened 2022-06-13
feat: a more comprehensive select component based on rc-select

xzdry:
A more comprehensive select component based on [rc-select ](https://github.com/react-component/select)         
The implementation references ant-design

https://user-images.githubusercontent.com/96771399/173327888-1a8a6796-8a73-4309-be51-fbc45160a264.mov

### DIFF
--- a/packages/editor-sdk/package.json
+++ b/packages/editor-sdk/package.json
@@ -38,6 +38,7 @@
     "lodash-es": "^4.17.21",
     "mitt": "^3.0.0",
     "mobx-react-lite": "^3.2.2",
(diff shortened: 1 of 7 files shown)


MrWindlike:
May use the `rc-select` in the `EnumField` too?

xzdry:
> May use the `rc-select` in the `EnumField` too?

Yes, I've tested it briefly, but I'm not sure if there are any other style problem. So for now only one part of the EventWidget has been modified. I will test it more thoroughly and replace all the chakra select

tanbowensg (approved):
LGTM

PR #453, opened 2022-06-14
feat(arco): enable to customize button content of Dropdown component

Mark-Fenng:
### motivation
Currently, the text of the Dropdown component is fixed with the text 'Click'.
image
image
This component isn't helpful unless the content is customisable.

### effect of PR
image
image


xzdry (approved):
It's great, I think it's also necessary to continue optimising the default components,I'll do it

Issue #455, opened 2022-06-15
Importing arco-lib causes app crash because 2 packages are missing.


tanbowensg:
**Version**

**Steps to reproduce**
Just create a new project, install sunmao-runtime and sunmao-arco-lib.

**What is expected?**
App runs correctly.

**What is actually happening?**
The app crashed, because 2 packages is missing.

![截屏2022-06-15 下午4 45 46](https://user-images.githubusercontent.com/12260952/173784566-0ad64e34-7566-4fd8-ba04-daafb9a84082.png)

Seems like arco-lib depends on editor-sdk and editor-sdk depends on chakra & mobx. As a result, arco-lib relies on chakra and mobx too although it should not.

tanbowensg:
fixed in https://github.com/webzard-io/sunmao-ui/pull/476

PR #457, opened 2022-06-16
update dependencies of packages

tanbowensg:
If you import arco-lib into a new project, it will fail to run, because 'framer-motion' and 'mobx' are missing. arco-lib depends on these packages indirectly, so users need to add them in project manually. I update packages' dependecies to solve this problem.
#455


tanbowensg:
still has some issues

PR #462, opened 2022-06-20
fix:  css style string is parsed to char array accidentally

Mark-Fenng:
### bug fix: CSS style string is parsed to char array accidentally

bug reproduction:
1. add CSS style 'padding: none;' to a style trait.
2. change the width or height value to '100px'.
image
3. Then the schema's corresponding part changes to 
image

#### motivation:
Currently, when a user clicks the table's radio or checkbox, the onRowClick method is also triggered.  In the general case, the row selection operation (clicking a checkbox) should be separated from clicking a row item. 
image
Thinking about a use case: a user clicks a row to open a model which shows detailed messages about it. The user doesn't want to open the dialogue when clicking the checkbox to select this row.


xzdry (commented):
Great

PR #464, opened 2022-06-21
feat(Link): add target property

xzdry:


packages/arco-lib/src/components/Link.tsx
```diff
@@ -12,8 +12,9 @@ const exampleProperties: Static<typeof LinkPropsSpec> = {
   disabled: false,
   hoverable: true,
   status: 'default',
-  href: 'https://www.smartx.com/',
+  href: '#',
   content: 'Link',
+  target: '_self',
```
tanbowensg, line 17: Maybe users don't know what does `_self` `_blank` mean. Can we translate `target` to boolean and rename it with a understandable name, like `OpenInNewPage`?

tanbowensg (changes requested):

packages/arco-lib/src/components/Link.tsx
Yuyz0112: 👍

xzdry: done

PR #471, opened 2022-06-22
fix(mask): fix the bug that mask location is wrong position when editor is scrolled

tanbowensg:
Fix it by adding scroll compensation to the mouse position.

https://user-images.githubusercontent.com/12260952/174958588-d3106f7e-4795-4e9e-9090-d55208552361.mov


Yuyz0112 (approved):
good catch

PR #473, opened 2022-06-22
feat: improve the typescript system

MrWindlike:
### feat: auto detect the raw properties types
- change the `properties` of the `ComponentSchema` and the `TraitSchema` to a specific type instead of `Record<string, unknown>`
- handle the condition where the raw properties are expressions

#### Updates
- add a `WidgetOptionsMap` interface to scale the widget type
- overwrite the `@sinclair/typebox`'s functions types to validate and promote the widget suggestions

##### How to expand the widget types?
When you implement a new widget in the other libs, you should expand the `WidgetOptionsMap` like this:

```js
const WidgetOptionsSpec = Type.Object({...});

declare module '@sunmao-ui/editor' {
  interface WidgetOptionsMap {
    'version/v1/custom': Static<typeof WidgetOptionsSpec>;
  }
}
```

##### Why define the `widget` and `widgetOptions` to `undefined` type in the `CustomOptions`?
1. If the `widget` is defined as `T extends keyof WidgetOptionsMap` and the `widgetOptions` as `WidgetOptionsMap[T]`, it would cause the cycle depending errors because the widget options also use the typebox to define.
2. In TypeScript, the function would match the best types depending on its parameters and return. If we define the `widget` as `string` type, when the widget type doesn't exist, it wouldn't report the errors anymore.

#### Preview

https://user-images.githubusercontent.com/25414733/175195120-7e4ea839-cfaf-471f-a824-7c772c4ae5fe.mov


packages/core/src/application.ts
```diff
@@ -1,6 +1,6 @@
 import { Metadata } from './metadata';
 import { parseVersion, Version } from './version';
-
+import { type PropsBeforeEvaled } from '@sunmao-ui/shared';
```
tanbowensg, line 3: This type had better to be in `core` rather than `shared`. Because if `shared` imported a type( which is very likely), there will be cycle dependency.

packages/shared/src/types/spec.ts
```diff
@@ -1,16 +1,15 @@
...
-  // category
-  category?: string;
-  weight?: number;
-  name?: string;
-  // conditional render
-  conditions?: Condition[];
-  // is a reference of component id
-  isComponentId?: boolean;
-};
+declare module '@sinclair/typebox' {
```
tanbowensg, line 3: This had better be in `runtime`, because not every lib imports `shared`.

tanbowensg, line 3: This had better be in `runtime`, because not every lib imports `shared`.

packages/arco-lib/src/generated/types/Cascader.ts
```diff
@@ -3,13 +3,16 @@ import { StringUnion } from '../../sunmao-helper';
 import { Category } from '../../constants/category';
 import { CORE_VERSION, CoreWidgetName } from '@sunmao-ui/editor-sdk';
 
-const EXPRESSION_WIDGET_TYPE = `${CORE_VERSION}/${CoreWidgetName.Expression}`;
+const EXPRESSION_WIDGET_TYPE = `${CORE_VERSION}/${CoreWidgetName.Expression}` as const;
```
tanbowensg, line 6: It is duplicated in many files.
And Is it a must to add `as const`?

tanbowensg (changes requested):

tanbowensg:
tests failed

packages/arco-lib/src/generated/types/Cascader.ts
MrWindlike: Yes. If it doesn't have the `as const`, it would be the `string` type instead of `core/v1/expression` type.

tanbowensg (approved):
LGTM

tanbowensg:
Conflict

PR #474, opened 2022-06-22
Use shallowCompare to compare slotProps

tanbowensg:
This can significantly improve the performance of runtime.


packages/shared/src/utils/shallowCompare.ts
```diff
@@ -0,0 +1,14 @@
+export function shallowCompare(obj1: any, obj2: any) {
```
MrWindlike, line 1: May moving it to `object.ts` is better? The utils files are classified by their service objects.

MrWindlike (approved):

PR #476, opened 2022-06-22
Make runtime ssr ready

tanbowensg:
Sunmao runtime is going to support ssr. The popular ssr frameworks like remix and nextjs don't support esm packges. So this pr replace or remove these esm packages.
This pr makes sure that the `runtime` is ssr ready. However, `arco-lib` and `chakra-ui-lib` are not, because they have some other packages which does not support ssr well.

### Changes
1. edit dependencies and peerDependencies
2. remove esm packages, like lodash-es...


Yuyz0112:
how about the bundle size

Yuyz0112 (approved):

PR #478, opened 2022-06-23
perf: improve the struct tree performance

MrWindlike:


packages/editor/src/components/StructureTree/StructureTree.tsx
```diff
@@ -30,6 +31,43 @@ const AutoCompleteStyle = css`
...
+
+    for (const i in realComponents) {
+      const preComponent = preRealComponents[i];
+      const component = realComponents[i];
+
+      if (preComponent !== component) {
+        // there are two situations would cause the ids are different
+        // 1. the component id is changed
+        // 2. the order of components is changed. example: [a, b] -> [b, a]
+        if (preComponent.id !== component.id) {
```
tanbowensg, line 52: I am not sure if the id check and slot check are necessary. Because the `components` is generated by `AppModel`. If a component has changeed, `AppModel` will change its object reference.

tanbowensg (changes requested):

packages/editor/src/components/StructureTree/StructureTree.tsx
MrWindlike: I think it is necessary. Because change the properties of the component or other traits would also change its object reference, it should separate this condition.

tanbowensg (approved):
LGTM

PR #479, opened 2022-06-23
fix: increase the hidden trait style priority

MrWindlike:
The hidden trait's `display: none` would be overwritten by the component styles or the style trait.

I use the `&&&& { display: none }` to increase its priority which is greater than the style trait's `&&&`.


tanbowensg (approved):
LGTM

PR #481, opened 2022-06-27
feat: use the array widget to render the event trait form

MrWindlike:
### Description
Currently, we use the tabs to render the event trait form which has some disadvantages:
- Can't see the handler information until you open the tab
- You need to open the tab before you change the order of the handlers

Thus, I use the array widget to replace the tabs.

### Preview
image

image


tanbowensg (approved):
LGTM

PR #482, opened 2022-06-27
perf: improve the struct tree performance when modifying

MrWindlike:
### Steps To Improve
1. Extract the components tree information into the store
2. Then get the needed slot information in the `ComponentTreeWrapper` and pass they to the `ComponentTree`
3. Compare the old and new `slotMap` in the `memo` to check whether the children update
4. Remove the `memoResolveApplicationComponents` and don't need to cache the tree information anymore because the updating only would rerender the related nodes now

#### Before
It takes 564.84ms to handle the drop event.
image

#### After
It takes 122.3ms to handle the drop event.
image


tanbowensg (approved):
LGTM

packages/editor/src/components/StructureTree/ComponentTree.tsx
```diff
@@ -3,48 +3,30 @@ import { Box, Text, VStack } from '@chakra-ui/react';
 import { ComponentSchema } from '@sunmao-ui/core';
 import { ComponentItemView } from './ComponentItemView';
 import { DropComponentWrapper } from './DropComponentWrapper';
-import { ChildrenMap } from './StructureTree';
 import { genOperation } from '../../operations';
 import { EditorServices } from '../../types';
 import { observer } from 'mobx-react-lite';
+import { isEqual } from 'lodash-es';
```
xzdry, line 9: `lodash-es` will have problems with ssr

MrWindlike: The SSR PR hasn't been merged into develop branch yet. I will change it later if the SSR PR is merged before this PR.

tanbowensg:
Confict

PR #486, opened 2022-06-30
fix: refreshing the StateView when switching between module and app

xzdry:


packages/editor/src/services/EditorStore.ts
```diff
@@ -89,6 +89,7 @@ export class EditorStore {
           this.setCurrentComponentsVersion(0);
           this.setLastSavedComponentsVersion(0);
           this.clearSunmaoGlobalState();
+          this.eventBus.send('stateRefresh');
           this.eventBus.send('componentsRefresh', this.originComponents);
```
tanbowensg, line 93: why not listen `componentsRefresh` event? They trigger at the same time.

tanbowensg (changes requested):

packages/editor/src/services/EditorStore.ts
xzdry: Because the `Editor` component is returned by the `initSunmaoUIEditor`, `componentsRefresh` will not trigger updates to the store in the Editor

xzdry: Because the `Editor` component is returned by the `initSunmaoUIEditor`, `componentsRefresh` will not trigger updates to the store in the Editor

packages/editor/src/init.tsx
```diff
@@ -110,6 +110,7 @@ export function initSunmaoUIEditor(props: SunmaoUIEditorProps = {}) {
       });
     }, []);
 
+    eventBus.on('stateRefresh', onRefresh);
```
tanbowensg, line 113: We should listen events in useEffect.

packages/editor/src/services/EditorStore.ts
tanbowensg: But if `componentsRefresh` can not work, why `stateRefresh` can? What's the diference?

tanbowensg (changes requested):

packages/editor/src/services/EditorStore.ts
xzdry: `componentsRefresh` does not trigger an update to the `_Editor` component view, but rather when `this.setComponents(newComponents)` changes the component
Because the `editorStore` is observed, the update of the view is triggered when the `components`  change. However, the `StateView` uses the `stateStore` created by `useState`:
```ts
const [store, setStore] = useState(stateManager.store);
```

and when switching between module and app, a new instance of store is created by calling `stateManager.clear()`:

```ts
  clear = () => {
    this.store = reactive<Record<string, any>>({});
  };
```
The original proxy for the store will no longer exist and `_Editor` component is still using the original `stateStore`, so no update will be triggered. Instead, the new instance will be set again in onRefresh with `setState(stateManager.store)`

packages/editor/src/init.tsx
xzdry: changed

tanbowensg (approved):
LGTM

PR #489, opened 2022-07-01
Fix/iframe

MrWindlike:
- use the `setInterval` to check whether the iframe is selected instead of listen to the window's blur event
- don't use the original `scrollIntoView` anymore because it would affect the parent page when the editor is inserted as an iframe

### DIFF
--- a/packages/editor/src/components/EditorMaskWrapper/EditorMaskWrapper.tsx
+++ b/packages/editor/src/components/EditorMaskWrapper/EditorMaskWrapper.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useRef, useState, useEffect, useCallback } from 'react';
+import React, { useMemo, useRef, useState } from 'react';
 import { EditorServices } from '../../types';
 import { observer } from 'mobx-react-lite';
 import { Box } from '@chakra-ui/react';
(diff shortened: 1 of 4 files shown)


packages/editor/src/components/EditorMaskWrapper/EditorMaskWrapper.tsx
```diff
@@ -66,25 +66,18 @@ export const EditorMaskWrapper: React.FC<Props> = observer(props => {
...
-    }
-  }, [setSelectedComponentId]);
-
   // can't capture the iframe click event
-  // use window's blur event to detect whether clicking the iframes
-  useEffect(() => {
-    window.focus();
-    window.addEventListener('blur', onClickIframe);
+  // use `setInterval` to check whether the iframe is selected
+  setInterval(function () {
```
tanbowensg, line 71: It should be in useEffect. And interval should be clear when unmount.

packages/editor/src/components/EditorMaskWrapper/EditorMaskWrapper.tsx
```diff
@@ -66,25 +66,18 @@ export const EditorMaskWrapper: React.FC<Props> = observer(props => {
...
 
-    return () => {
-      window.removeEventListener('blur', onClickIframe);
-    };
-  }, [onClickIframe]);
+      if (selectedComponentId !== currentSelectedComponentId) {
+        setSelectedComponentId(currentSelectedComponentId);
+      }
+    }
+  }, 1000);
```
tanbowensg, line 80: 1000ms is too long, 100ms will be better.

tanbowensg (changes requested):

tanbowensg (approved):

PR #493, opened 2022-07-04
refactor: refactor parserTypebox

xzdry:
Refactor the `parserTypebox` function to get the initial value of the schema based on spec.type


packages/shared/src/utils/spec.ts
```diff
@@ -31,37 +22,80 @@ export function StringUnion<T extends string[]>(values: [...T], options?: any) {
...
+    } else {
+      obj[key] = parseTypeBox(subSpec);
+    }
+  }
+  return obj;
+}
+
+export function parseTypeBox(spec: JSONSchema7): JSONSchema7Type {
+  // compatible with typebox optional
+  if ((spec as TSchema).modifier === OptionalModifier) {
```
MrWindlike, line 52: May check the optional property by object type schema's `required` properties in the `getObject` function instead of using the `typebox`'s `OptionalModifier`?

packages/shared/src/utils/spec.ts
```diff
@@ -31,37 +22,80 @@ export function StringUnion<T extends string[]>(values: [...T], options?: any) {
...
+    if (typeof subSpec === 'boolean') {
+      obj[key] = null;
+    } else {
+      obj[key] = parseTypeBox(subSpec);
+    }
+  }
+  return obj;
+}
+
+export function parseTypeBox(spec: JSONSchema7): JSONSchema7Type {
```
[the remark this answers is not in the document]
Yuyz0112: +1

parse usually means parse raw format to a struct, like `JSON.parse`. This function act as `generate/mock/factory`

packages/shared/src/utils/spec.ts
xzdry: Good idea, I'll change it

xzdry: done

packages/shared/src/utils/spec.ts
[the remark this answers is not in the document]
xzdry: changed

tanbowensg (approved):

PR #499, opened 2022-07-07
feat: make the slot more compact

MrWindlike:
This PR removes the "Empty" text from the slot to make it more compact. 

And now you can drop the component into the slot content whether or not it has children which would append to the front of the slot.

image

### DIFF
--- a/packages/editor/src/AppModel/ComponentModel.ts
+++ b/packages/editor/src/AppModel/ComponentModel.ts
@@ -169,15 +169,12 @@ export class ComponentModel implements IComponentModel {
     return trait;
   }
 
-  appendTo = (parent?: IComponentModel, slot?: SlotName) => {
+  appendToSlotFront = (parent: IComponentModel, slot: SlotName) => {
     // remove from current position
     if (this.parent) {
       this.parent.removeChild(this);
     }
-    if (!parent || !slot) {
-      this.appModel.appendChild(this);
-      return;
-    }
+
     // update parent
     if (!parent.children[slot]) {
       parent.children[slot] = [];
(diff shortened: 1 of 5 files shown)


packages/editor/src/AppModel/ComponentModel.ts
```diff
@@ -183,7 +183,7 @@ export class ComponentModel implements IComponentModel {
       parent.children[slot] = [];
     }
 
-    parent.children[slot].push(this);
+    parent.children[slot].unshift(this);
```
tanbowensg, line 186: Why use unshift? `append` usually means add to the last position.
If you want add a component before other siblings, you should first call `appendTo`, then call `moveAfter`.

tanbowensg (changes requested):

packages/editor/src/AppModel/ComponentModel.ts
MrWindlike: I want to append it to the front of the slot when dropping a component into the slot item. I split the `appendTo` into `appendToRoot` and `appendToSlotFront` two functions. What do you think of this approach?

tanbowensg: Maybe it is better to add component at the first in dropping situation, but in other situation it is not. For example, user drop component into component element in canvas. Actually, we cannot prediect whether user wants to append component to the first or the last.
There is no way to satisfy all the situations. So I think we just keep it simple and make no changes.

MrWindlike: Done. I revert the related codes.

tanbowensg (approved):

PR #502, opened 2022-07-07
fix: change the logic of the expression transform

MrWindlike:
## Changes
When the value passing to the expression widget isn't the string type, it would be transformed by wrapping `{{}}` to it.

Submitting the expression code would parse the expression code by the spec type instead of the original value type.

## Why
The current implementation would use the original value type to parse the expression code which couldn't change the value type. For example, if I have a property with the `Type.Any()` type and its value is an object, I want to change it to the string type, but I can't do it.


tanbowensg (approved):
LGTM. But I'am not sure that it is good to auto add `{{}}`. We merge it for now, and see how it will work.

PR #503, opened 2022-07-08
Refactor/arco

xzdry:
1. Skeleton: remove union type
2. PasswordInput: add default value
3. Input/passwordInput: fix a bug where inserting a slot causes it to be unselected(Cannot get the latest dom element by `ref.current`)
4. others: Add missing titles, improve descriptions, re-categorise and use appropriate widgets for display


tanbowensg (approved):
LGTM

PR #504, opened 2022-07-08
feat(arco/Slider): add formatTooltip slot

xzdry:
image
The tooltip can now be formatted with formatTooltip slot

### DIFF
--- a/packages/arco-lib/src/components/Slider.tsx
+++ b/packages/arco-lib/src/components/Slider.tsx
@@ -37,13 +37,19 @@ export const Slider = implementRuntimeComponent({
     properties: SliderPropsSpec,
     state: SliderStateSpec,
     methods: {},
-    slots: {},
+    slots: {
+      tooltip: {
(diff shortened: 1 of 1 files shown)


tanbowensg:
I doubt whether it is necessary to add slot for tooltip. What do you want to put in tooltip's slot?

xzdry:
> I doubt whether it is necessary to add slot for tooltip. What do you want to put in tooltip's slot?

tanbowensg (approved):
LGTM

tanbowensg:
OK. But I think the slot name should be `tooltip`, not `formatTooltip`.

tanbowensg (changes requested):
OK. But I think the slot name should be tooltip, not formatTooltip.

xzdry:
> OK. But I think the slot name should be tooltip, not formatTooltip.

Sounds better, have changed

tanbowensg (approved):

PR #505, opened 2022-07-08
fix(arco/Menu): fix a bug where the collapse button doesn't work

xzdry:
When `hasCollapseButton` is false, it is in `uncontrolled mode` and the collapse button is clicked to collapse the menu, when it is true, the `collapse property` controls whether it is collapsed or not.


tanbowensg (approved):
LGTM

PR #508, opened 2022-07-11
feat: refactor the validation trait (#breaking-chagnes)

MrWindlike:
### Changes
- Support more built-in validate functions
- Support custom the validate function
- Support multiple validation rules for a value
- Triggered the validation by methods instead of the value change
- Change the spec struct(breaking changes)


tanbowensg (approved):
The spec seems ok. LGTM.

PR #515, opened 2022-07-12
feat(runtime): pass slot fallback into slot elements

Yuyz0112:
Previously, we introduce if condition to the slot trait, which helps
render slot element conditionally. But it is not easy for a component
developer to implement fallback UI of a slot, because neither the
if condition is true or false, the slot elements function will return
the ImplWrapperMain component.

This patch brings a new way to implement fallback UI: pass the
fallback UI as the second parameter of the slot elements function.
So in the ImplWrapper, we can fall back to the passed in UI.

Before:
image

After:
image

### DIFF
--- a/packages/runtime/src/components/_internal/ImplWrapper/UnmountImplWrapper.tsx
+++ b/packages/runtime/src/components/_internal/ImplWrapper/UnmountImplWrapper.tsx
@@ -8,7 +8,7 @@ import { watch } from '../../..';
 
 export const UnmountImplWrapper = React.forwardRef<HTMLDivElement, ImplWrapperProps>(
   function UnmountImplWrapper(props, ref) {
-    const { component: c, services } = props;
+    const { component: c, services, slotFallback } = props;
     const { stateManager, registry } = services;
     const { executeTrait } = useRuntimeFunctions(props);
 
@@ -76,6 +76,10 @@ export const UnmountImplWrapper = React.forwardRef<HTMLDivElement, ImplWrapperPr
       initSingleComponentState(registry, stateManager, c);
     }
 
-    return !isHidden ? <ImplWrapperMain {...props} ref={ref} /> : null;
+    return !isHidden ? (
+      <ImplWrapperMain {...props} ref={ref} />
+    ) : slotFallback ? (
+      <>{slotFallback}</>
+    ) : null;
   }
 );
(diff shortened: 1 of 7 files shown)


Yuyz0112:
Currently I only migrate the Tabs' code to the new fallback way, if this works, we can migrate other slot elements to this way,

packages/runtime/src/components/_internal/ImplWrapper/UnmountImplWrapper.tsx
```diff
@@ -76,6 +76,10 @@ export const UnmountImplWrapper = React.forwardRef<HTMLDivElement, ImplWrapperPr
       initSingleComponentState(registry, stateManager, c);
     }
 
-    return !isHidden ? <ImplWrapperMain {...props} ref={ref} /> : null;
+    return !isHidden ? (
+      <ImplWrapperMain {...props} ref={ref} />
+    ) : slotFallback ? (
+      <>{slotFallback}</>
+    ) : null;
```
tanbowensg, line 83: If there are multiple hidden children, will it render multiple placeholders?

tanbowensg: Yes, but a slot can contain multilple components.
截屏2022-07-13 下午6 16 31

Yuyz0112: good catch!

Just update the implementation, now use a render set to check whether all the slot elements are hidden and only attach fallback to the last hidden slot.

packages/runtime/src/components/_internal/ImplWrapper/hooks/useSlotChildren.tsx
```diff
@@ -17,8 +17,20 @@ export function getSlotElements(
       return <ImplWrapper key={child.id} {...props} component={child} />;
     });
 
-    slotElements[slot] = function getSlot(slotProps) {
-      return slotChildren.map(child => React.cloneElement(child, { slotProps }));
+    slotElements[slot] = function getSlot(slotProps, slotFallback) {
+      const slotContext = {
+        renderSet: new Set(childrenMap[c.id][slot].map(child => child.id)),
```
tanbowensg, line 22: Why do we need Set here? Every component's id is unique. It can not be duplicated.

Yuyz0112: Because we need to remove the id when it's hidden and check the remaining size, a set is a little easier and more efficient to remove an element than an array.

I've also tried to use a number as the counter, but some React double render mechanisms make it very unstable.

tanbowensg (approved):
LGTM

PR #516, opened 2022-07-13
fix(runtime): remove react-markdown

tanbowensg:
Remove react-markdown because its 6.x version will cause vite build error and 7.x+ version only export esm package which blocks ssr.


Yuyz0112:
LGTM, and this introduce a breaking change?

tanbowensg:
Yes, but i checked in installer and zbs-insight. They don't use `md` format text.

Yuyz0112 (approved):

PR #528, opened 2022-07-15
refactor(core/Stack): The spacing of the stack is changed to set the margin of the child element for better compatibility

xzdry:


packages/runtime/src/components/core/Stack.tsx
```diff
@@ -65,7 +86,18 @@ const Stack = React.forwardRef<HTMLDivElement, StackProps & { children: ReactNod
           ${cssStyle}
         `}
       >
-        {children}
+        {childs.map((child, idx) => {
+          return (
+            <div
```
tanbowensg, line 91: Wrappering with div will break the structure and make children's some css not working, like `flex: 1 1 auto`... 
Can we use css selector to add style to children? like `:last-child `

tanbowensg (changes requested):

Yuyz0112:
Maybe we can check some other CSS/UI frameworks about how they handle the compatibility of row-gap/column-gap.

PR #529, opened 2022-07-18
support for automatic generation of `rowKey` or setting from columns

xzdry:
Support for specifying a column as a row key and automatic generation of a row key by default
image

### DIFF
--- a/packages/arco-lib/src/components/Table/Table.tsx
+++ b/packages/arco-lib/src/components/Table/Table.tsx
@@ -72,10 +72,16 @@ const rowClickStyle = css`
   }
 `;
 
-const tableRowKey = Symbol.for('table.rowKey');
-
 export const exampleProperties: Static<typeof TablePropsSpec> = {
   columns: [
+    {
+      title: 'Key',
+      dataIndex: 'key',
+      type: 'text',
+      displayValue: '',
+      filter: false,
+      componentSlotIndex: 0,
+    },
     {
       title: 'Name',
       dataIndex: 'name',
@@ -109,11 +115,12 @@ export const exampleProperties: Static<typeof TablePropsSpec> = {
   data: Array(13)
     .fill('')
     .map((_, index) => ({
+      key: index,
       name: `${Math.random() > 0.5 ? 'Kevin Sandra' : 'Naomi Cook'}${index}`,
       link: `link${Math.random() > 0.5 ? '-A' : '-B'}`,
       salary: Math.floor(Math.random() * 1000),
     })),
-  rowKey: 'auto',
+  rowKey: 'key',
   checkCrossPage: true,
   pagination: {
     enablePagination: true,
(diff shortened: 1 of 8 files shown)


packages/arco-lib/src/widgets/TablePrimaryKeyWidget.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const TablePrimaryKeyWidget: React.FC<
+  WidgetProps<TablePrimaryKeyWidget, string>
+> = props => {
+  const { value, onChange, component } = props;
+  const { properties } = component;
+  const columns = properties.columns as Static<typeof ColumnSpec>[];
+
+  const keys = ['auto', ...columns.map(c => c.dataIndex)];
```
tanbowensg, line 23: Dose `auto` means a key called `auto` or auto choosing a key? I could be conflict.

packages/arco-lib/src/components/Table/Table.tsx
```diff
@@ -196,6 +206,7 @@ export const Table = implementRuntimeComponent({
 
   useEffect(() => {
     mergeState({ pageSize });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tanbowensg, line 209: Can we make sure the `pageSize` will never change or don't need to mergeState when it changes?

packages/arco-lib/src/components/Table/Table.tsx
```diff
@@ -231,13 +242,39 @@ export const Table = implementRuntimeComponent({
...
       // If the `useCustomPagination` is true, then no pagination will be done and data over the pagesize will be sliced
       // Otherwise it will automatically paginate on the front end based on the current page
-      return sortedData?.slice(0, pageSize);
+      data = sortedData?.slice(0, pageSize);
     }
-    return sortedData;
-  }, [pageSize, sortedData, enablePagination, useCustomPagination]);
+
+    // auto-generated row key
+    return rowKey === 'auto'
```
tanbowensg, line 253: Do we need to auto generate key for users? I think it rarely happens.

tanbowensg (changes requested):

packages/arco-lib/src/widgets/TablePrimaryKeyWidget.tsx
xzdry: auto means add a key to each item of data

packages/arco-lib/src/components/Table/Table.tsx
xzdry: If a default column is not unique as a rowKey, this can also lead to duplicate selections.

tanbowensg: Yes, but I think it should leave to users to decide which is the `rowKey` and make sure it is unique.
Users don't know the rule of auto generating `rowKey`, so even we generated it for user, they won't know what dost it look like.

packages/arco-lib/src/components/Table/Table.tsx
xzdry: This is only used to synchronise the initial state of pageSize

packages/arco-lib/src/components/Table/Table.tsx
xzdry: Removed

tanbowensg (approved):
LGTM

PR #532, opened 2022-07-18
fix: avoid the expression tip would be covered

MrWindlike:


tanbowensg:
should rebase with main branch.

PR #536, opened 2022-07-19
feat: add `doOperations`

MrWindlike:
Allow the widgets do some operations.


packages/editor/src/init.tsx
```diff
@@ -100,6 +113,7 @@ export function initSunmaoUIEditor(props: SunmaoUIEditorProps = {}) {
     widgetManager,
     eventBus,
     editorStore,
+    doOperations,
```
tanbowensg, line 116: Can we move this function to `appModelManager`? Because this function is quiet different with other services.

tanbowensg (changes requested):

packages/editor/src/init.tsx
MrWindlike: Done.

PR #537, opened 2022-07-19
fix(runtime): only teardown state when the component is not hidden before this check

Yuyz0112:
when we are using slot's ifCondition, the slot may have
unmount traits result multiple times


packages/runtime/__tests__/ImplWrapper/ImplWrapper.spec.tsx
```diff
@@ -132,3 +133,46 @@ describe('when component merge state asynchronously', () => {
...
+        },
+      }
+    `);
+    act(() => {
+      screen.getByTestId('tabs-tab-1').click();
+    });
+    expect(stateManager.store).toMatchInlineSnapshot(`
+      Object {
+        "input": Object {
+          "value": "new-value",
```
Yuyz0112, line 167: Without the fix, here will be `value: ""`, because the state has been teardown and re-init.

tanbowensg (approved):
LGTM

PR #542, opened 2022-07-22
fix(structureTree): if a component item is already in viewport, then don't scroll it into view

tanbowensg:
This pr can fix the problem that the component tree scrolls too frequently.

### DIFF
--- a/packages/editor/src/components/StructureTree/StructureTree.tsx
+++ b/packages/editor/src/components/StructureTree/StructureTree.tsx
@@ -35,8 +35,7 @@ export const StructureTree: React.FC<Props> = props => {
   const [search, setSearch] = useState('');
   const { components, onSelectComponent, services } = props;
   const { editorStore } = services;
-  const wrapperRef = useRef<HTMLDivElement>(null);
-
+  const scrollWrapper = useRef<HTMLDivElement>(null);
   const onSelectOption = useCallback(
     ({ item }: { item: Item }) => {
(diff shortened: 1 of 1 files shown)


packages/editor/src/components/StructureTree/StructureTree.tsx
```diff
@@ -49,10 +48,19 @@ export const StructureTree: React.FC<Props> = props => {
       // wait the component tree to be expanded
       setTimeout(() => {
         const selectedElement: HTMLElement | undefined | null =
-          wrapperRef.current?.querySelector(`#tree-item-${selectedId}`);
+          scrollWrapper.current?.querySelector(`#tree-item-${selectedId}`);
 
-        if (selectedElement) {
-          scrollIntoView(selectedElement, { time: 0, align: { lockX: true } });
+        const wrapperRect = scrollWrapper.current?.getBoundingClientRect();
```
MrWindlike, line 53: May cache the wrapper position information instead of re-reading it every time?

tanbowensg: I think it is not necessary. Because it only runs when being clicked. And the wrapper size will change if user resize it or browser. So there is no benefits to cache it and will bring more problems.

MrWindlike: OK.

MrWindlike (approved):

PR #543, opened 2022-07-22
fix(shared): use default value instead of  undefined when generate default value from spec

tanbowensg:
# Problem
The optional properties' field of newly created component/trait does not show in form. For example, the hidden trait has a `visually` property and it does not show.
截屏2022-07-22 下午6 16 50

This is because the form is rendered from the `defaultValue` generated from spec. And the `generateDefaultValueFromSpec` function will not generate value if a property is optional. 

# Solution

So I changed the logic of `generateDefaultValueFromSpec`. It will generate default value for optional property now.


xzdry:
LGTM

xzdry (approved):

PR #547, opened 2022-07-27
fix(ImplWrapper): deep compare slotContext in memo

tanbowensg:
# Problem
Every time a parent component updates, its `slotContext.renderSet` will be a new `Set`. This causes `shallowCompare` fail to check two same `slotContext`, leading to the fact that every time a parent updates, its children will update too, no matter its children changes or not.

I replace `shallowCompare` to `isEuqual` to deep compare to change. Since `slotContext` is not very big, I think the trade off of performance is acceptable.

截屏2022-07-27 下午3 50 58


Yuyz0112 (approved):
good catch

PR #548, opened 2022-07-27
chore: reexport the shared package by runtime

MrWindlike:


Yuyz0112 (approved):
BTW, maybe we can stop publish internal packages like `@sunmao-ui/shared` to npm?

PR #552, opened 2022-07-29
feat(RecordWidget): allow use the widget for object

MrWindlike:
Let the `RecordWidget` use other widgets to display the object and array type properties when explicitly declaring the widget in the spec. 

When the spec is `Type.Record(Type.String(), Type.Object(...))`, generating the default value by the spec which read from `spec.patternProperties.['^.*$']` instead of using the empty string as the default value.

### DIFF
--- a/packages/editor-sdk/src/components/Form/RecordEditor.tsx
+++ b/packages/editor-sdk/src/components/Form/RecordEditor.tsx
@@ -10,6 +10,8 @@ import { ExpressionEditor } from '../Form/ExpressionEditor';
 import { WidgetProps } from '../../types/widget';
 import { mergeWidgetOptionsIntoSpec } from '../../utils/widget';
 import { ExpressionEditorProps } from './ExpressionEditor';
+import { generateDefaultValueFromSpec } from '@sunmao-ui/shared';
+import { JSONSchema7 } from 'json-schema';
 
 const IGNORE_SPEC_TYPES = ['array', 'object'];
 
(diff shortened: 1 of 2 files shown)


packages/editor-sdk/src/components/Form/RecordEditor.tsx
```diff
@@ -60,16 +62,16 @@ const RowItem = (props: RowItemProps) => {
     }),
     []
   );
-  const valueSpec = useMemo(() => spec?.properties?.[rowKey], [spec, rowKey]);
+  const valueSpec = useMemo(
+    () => (onlySetValue ? spec?.properties?.[rowKey] : spec?.patternProperties?.['^.*$']),
```
Yuyz0112, line 66: does all the spec has `patternProperties`?

MrWindlike: No. When the spec is `Type.Record(Type.String(), Type.Any())`, it has the `patternProperties`. But it doesn't matter if it doesn't have the `patternProperties`, because the value's input widget would be the expression widget.

Yuyz0112: I mean is this a typebox implementation or a general JSON-schema implementation?

MrWindlike: The `patternProperties` is a general JSON-schema implementation.

image

Yuyz0112 (approved):

PR #553, opened 2022-07-29
use the new data path to pass slot props and fallback elements

Yuyz0112:
Overview:

image

The main idea of this patch is to set slot props to a reactive store instead of passing them as React props to the slot children elements. On the other hand, we put fallback elements into a Map either, to avoid of passing them as props.

## Pass slot props

Same as other state management in sunmao, the component will subscribe to the slot store to get the latest value of slot props.

Same as component id, we need a way to identify every slot prop in the store. So we add a new parameter key to the `getSlot` function, where [component id, slot name, key] will be a unique identifier.

> Currently, the PR you are reviewing only migrates several components with the new `key` parameter, but we will migrate all if the solution passes the review.

Another challenge is to subscribe to the slot store in an efficient way. Currently, we are using Vue's reactivity system, which needs to set `deep: true` to observe multiple values during effect.

By setting `deep: true`, if we write code like `const $slot = slotStore[slotKey]`, it will subscribe to all the sub-fields of this slot key because of we 'access' the property.

The solution is to implement a redirector Proxy. With the redirector, expression `$slot.a.b` will be redirect to `slotStore[slotKey].a.b`.

## Pass slot fallback elements

The slot receiver implementation is a magic hole of sunmao's runtime.

In sunmao, we support pass props and fallback elements to a slot. But if we pass them as React component's props, it will cause re-render since most of them could not use a shallow equal checker.

Also, in sunmao's runtime, we **are not using** a traditional React render mechanism. Instead, we keep most of the components not rendered and only subscribe to related state updates.

To continue with our design, we need a way to render the slot's fallback elements without passing the elements as props.

This is where the slot receiver comes. It contains a map and an event emitter, when a slot needs to render, it will attach the fallback elements to the map and send a signal via the emitter.

When the Receiver component receives a signal, it will **force render** the fallback elements.

## Benefits

With the new approach, our `ImplWrapper` becomes very clean again, which does not need to check the equality of slot props and slot context.

By introducing the `key` concept, it may solve the conflict of id in loop-render slots.

Also, I suggest unifying the previous `List` implementation with this approach, which I've done a simple demo in this patch.

### DIFF
--- a/packages/runtime/src/components/_internal/ImplWrapper/SlotReciver.tsx
+++ b/packages/runtime/src/components/_internal/ImplWrapper/SlotReciver.tsx
@@ -1,0 +1,64 @@
+/**
+ * The slot receiver is a magic hole of sunmao's runtime.
+ * In sunmao, we support pass props and fallback elements to a slot.
+ * But if we pass them as React component's props,
+ * it will cause re-render since most of them could not use a shallow equal checker.
+ *
+ * Also, in sunmao's runtime, we are not using a traditional React render mechanism.
+ * Instead, we keep most of the components not be rendered and only subscribe to related state updates.
+ *
+ * To continue with our design,
+ * we need a way to render slot's fallback elements without passing the elements as props.
+ * This is where the slot receiver comes.
+ * It contains a map and an event emitter, when a slot need render,
+ * it will attach the fallback elements to the map and send a signal via the emitter.
+ * When the Receiver component receive a signal, it will force render the fallback elements.
+ */
+import React, { useEffect, useState, useRef } from 'react';
+import mitt from 'mitt';
+
+class SlotReceiver {
+  fallbacks: Partial<Record<string, React.ReactNode>> = {};
+  emitter = mitt<Record<string, React.ReactNode>>();
+
+  constructor() {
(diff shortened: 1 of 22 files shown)


packages/runtime/src/components/_internal/ImplWrapper/SlotReciver.tsx
```diff
@@ -0,0 +1,64 @@
...
+      // undefined means no fallback has been received yet
+      // null means the Receiver component start to hanle the events
+      if (this.fallbacks[slotKey] === undefined) {
+        this.fallbacks[slotKey] = c;
+      }
+    });
+  }
+}
+
+export const slotReceiver = new SlotReceiver();
```
tanbowensg, line 35: We need move it in to `services`.

Yuyz0112: 👍

Yuyz0112: done

Yuyz0112:
rebased the latest upstream

tanbowensg (approved):
LGTM

PR #555, opened 2022-07-29
fix(editor): fix a bug which will cause the mask of chakra-ui/checkbo…

FriedRiceNoodles:
…x render incorrectly

Chakra-ui/Checkbox set its ref in an inner 1px * 1px input element, so the EditorMaskManager got a 1px*1px hoverComponent and selectedComponent, then render a 1px * 1px mask.
Chakra-ui/Checkbox把它ref透传到了它内部的一个input元素中，而这个元素的宽高只有1px，所以在编辑器中这个checkbox无法被hover到，且选中时渲染出来的mask也只有1px的宽高。
image
image

This is a temporary fix, in future, i will push a feature to separate elementRef and mask, it may look like:
这是一次临时性修复，我会想把控制行为的elementRef和mask分离开来，而不是通过这样给某个组件打补丁的方式来解决这类问题。
image

### DIFF
--- a/packages/chakra-ui-lib/src/components/Checkbox.tsx
+++ b/packages/chakra-ui-lib/src/components/Checkbox.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import { Static, Type } from '@sinclair/typebox';
 import { Checkbox as BaseCheckbox, useCheckboxGroupContext } from '@chakra-ui/react';
 import { implementRuntimeComponent, Text, TextPropertySpec } from '@sunmao-ui/runtime';
@@ -116,7 +116,7 @@ export default implementRuntimeComponent({
     colorScheme,
     mergeState,
     customStyle,
-    elementRef,
+    getElement,
   }) => {
     const groupContext = useCheckboxGroupContext();
     let _defaultIsChecked = false;
@@ -126,6 +126,8 @@ export default implementRuntimeComponent({
       _defaultIsChecked = groupContext.value.some(val => val === value);
     }
     const [checked, setChecked] = useState(_defaultIsChecked);
+
+    const ref = useRef<HTMLInputElement>(null);
 
     useEffect(() => {
       mergeState({ text: text.raw });
@@ -142,6 +144,12 @@ export default implementRuntimeComponent({
     useEffect(() => {
       setChecked(!!defaultIsChecked);
     }, [setChecked, defaultIsChecked]);
+
+    useEffect(() => {
+      if (getElement && ref.current) {
+        getElement(ref.current.parentElement as HTMLElement);
+      }
+    });
 
     const args: {
       colorScheme?: Static<ReturnType<typeof getColorSchemePropertySpec>>;
@@ -170,7 +178,7 @@ export default implementRuntimeComponent({
         className={css`
           ${customStyle?.content}
         `}
-        ref={elementRef}
+        ref={ref}
       >
         <Text value={text} />
       </BaseCheckbox>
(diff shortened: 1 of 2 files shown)


FriedRiceNoodles:
BTW, how can i invite or request a reviewer?(I'm new to GitHub.)

Yuyz0112:
@FriedRiceNoodles Thanks for the PR! I've added the maintainers of chakra-ui lib as reviewers, they will review the PR after the weekend.

tanbowensg:
Perhaps we can not merge this pr. Adding code in `EditorMaskManager` is not the correct way to fix this bug. Chakra is just a lib of sunmao, while `EditorMaskManager` is part of `Editor`, and the `Editor` should not be coupled with particular lib.

The correct way to fix this problem is to pass the correct ref to sunmao by `elementRef` or `getElement`. If chakra-ui checkbox doesn't expose its wrapper's ref, then wrap it with a `div` maybe the only solution.

FriedRiceNoodles:
You are right. I have considered what you said, but i need to make sure that only the mask depends on `elementRef`, or we may get new bugs by wrapping checkbox with a `div` and set the `ref` to outer div.

tanbowensg:
You can pass a special `classname` to chakra checbox and Use DOM API to get the label element of checkbox, then pass the element to `getElement`.

FriedRiceNoodles:
Well, I just did what you said, but used `parentElement` instead of a special `className`.

tanbowensg (approved):
LGTM

packages/chakra-ui-lib/src/components/Checkbox.tsx
```diff
@@ -143,6 +145,12 @@ export default implementRuntimeComponent({
       setChecked(!!defaultIsChecked);
     }, [setChecked, defaultIsChecked]);
 
+    useEffect(() => {
+      if (getElement && ref.current) {
+        getElement(ref.current.parentElement as HTMLElement);
```
Yuyz0112, line 150: @tanbowensg 
Do we need to call `getElement` in every render call? This `useEffect` hook does not have a dependency array.

FriedRiceNoodles: Here is a mistake, it should be:
```javascript
    useEffect(() => {
      if (getElement && ref.current) {
        getElement(ref.current.parentElement as HTMLElement);
      }
    }, [getElement, ref]);
```

Yuyz0112: I think the ref does not need to be added to the dependency.

https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/

FriedRiceNoodles:
I copied it from other components&nbsp; and docs/zh/components.md



------------------&nbsp;原始邮件&nbsp;------------------
发件人: ***@***.***&gt;; 
发送时间: 2022年8月2日(星期二) 下午2:00
收件人: ***@***.***&gt;; 
抄送: ***@***.***&gt;; ***@***.***&gt;; 
主题: Re: [smartxworks/sunmao-ui] fix(editor): fix a bug which will cause the mask of chakra-ui/checkbo… (PR #555)





 
@Yuyz0112 commented on this pull request.
 
 
In packages/chakra-ui-lib/src/components/Checkbox.tsx:
 &gt; @@ -143,6 +145,12 @@ export default implementRuntimeComponent({        setChecked(!!defaultIsChecked);      }, [setChecked, defaultIsChecked]);   +    useEffect(() =&gt; { +      if (getElement &amp;&amp; ref.current) { +        getElement(ref.current.parentElement as HTMLElement);  
I think the ref does not need to be added to the dependency.
 
https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/

FriedRiceNoodles:
I just read the article you sent, then I search '[getElement, ref]' in sunmao and found this.
image
Should we remove all the `ref` in dependency arrays？

PR #556, opened 2022-08-01
refactor: let the `deepEval` support eval the expression string

MrWindlike:


packages/runtime/src/services/StateManager.ts
```diff
@@ -157,58 +164,67 @@ export class StateManager {
...
-          return result;
-        },
-        newV => {
-          if (isProxy(newV)) {
-            newV = toRaw(newV);
+    const evaluated = this.deepEval(value, options) as T extends string
+      ? unknown
+      : PropsAfterEvaled<Exclude<T, string>>;
+
+    if (value && typeof value === 'object') {
```
tanbowensg, line 198: If value is string, I think maybe we also need to watch it?

tanbowensg (changes requested):

packages/runtime/src/services/StateManager.ts
MrWindlike: Done.

tanbowensg (approved):
LGTM

PR #559, opened 2022-08-02
Refactor some expressions logics

tanbowensg:
# Principles
1. When switching from expression mode, editor should immediately save once.
2. Converting data types should not be done by `stateManager`, but by the widget.
3. Each Widget should convert the incoming value , and if the widget can not render value, it should report an error, not modify the schema.
4. ExpressionWidget should not convert the user input string.


Yuyz0112 (approved):
LGTM

PR #567, opened 2022-08-11
fix: fix clicking outside can't close the popover widget

MrWindlike:
when clicking the outside node which would stop propagation, it can't close the popover widget


tanbowensg (approved):
LGTM

PR #568, opened 2022-08-15
fix: fix the `exampleProperties` type errors

MrWindlike:
The previous approach causes the 'exampleProperties' error when the component property is Record type.

Because the type of `T & {
    metadata: { exampleProperties: DeepPartial<Static<T['spec']['properties']>> };
  }` is a hack way to validate the `exampleProperties`, it would cause errors in some situations. 

image

image

This PR uses the generics type as the `exampleProperties` type, then fixes the errors.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -37,7 +37,6 @@
   "devDependencies": {
     "@swc/core": "^1.2.121",
     "@sinclair/typebox": "^0.21.2",
-    "@sunmao-ui/shared": "^0.2.3",
     "@types/jest": "^26.0.23",
     "@types/json-schema": "^7.0.7",
     "@types/lodash": "^4.14.182",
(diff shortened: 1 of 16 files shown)


packages/core/package.json
```diff
@@ -36,6 +36,8 @@
   },
   "devDependencies": {
     "@swc/core": "^1.2.121",
+    "@sinclair/typebox": "^0.21.2",
+    "@sunmao-ui/shared": "^0.2.3",
```
tanbowensg, line 40: core should not depend on shared.

tanbowensg (changes requested):

packages/core/package.json
MrWindlike: OK.

tanbowensg (approved):
LGTM

PR #569, opened 2022-08-15
Split arco-lib's and chakra-lib's lib and widgets to two entries

tanbowensg:
# Background
Arco-lib's lib and widgets are both in one package. If a project imports arco-lib, both of them will be built into the runtime bundle.

So I add split `lib` and `widgets` to two entries. When building runtime app, user can only import `'arco-lib/lib'`. This will not import widgets and the dependencies of widgets.


packages/arco-lib/package.json
```diff
@@ -11,21 +11,19 @@
...
   "scripts": {
     "dev": "vite",
-    "typings": "tsc --emitDeclarationOnly",
-    "build": "tsup src/index.ts src/widgets/index.ts --format cjs,esm,iife --legacy-output --inject ./react-import.js --clean --no-splitting --sourcemap",
+    "build": "tsup",
     "serve": "vite preview",
     "lint": "eslint ./src --ext .ts --ext .tsx",
     "fix-lint": "eslint --fix ./src --ext .ts --ext .tsx",
-    "prepublish": "npm run build && npm run typings"
+    "prepublish": "npm run build"
```
Yuyz0112, line 26: I think previously we had some issue with tsup's dts generation, have you tried to remove all the typings files first and use tsup to generate it again?

tanbowensg:
Close because I have changed a lot of this pr.

PR #573, opened 2022-08-17
fix: fix selecting component would change the properties' values issue

MrWindlike:
## Issue
When selecting a component, it would change the properties' values.

### Repetition steps
1. Refresh
2. Select the component whose `ifCondition` doesn't use the expression
3. Select another component whose `ifCondition` uses the expression
4. You can see the expression would be a switch component, and the `ifCondition` value is `true`

https://user-images.githubusercontent.com/25414733/185020448-99f30db1-3e56-47b9-8739-283a5af2f5a5.mov

## Reason
The different components would reuse the form, when selecting another component it would trigger the widget's `useEffect` and then would change the value.

image

## Fix
Add the different keys for the different component forms.


tanbowensg (approved):
LGTM

PR #586, opened 2022-09-01
add slot key

xzdry:
1. add slot key to keep  that `$slot` gets the correct value     
2. add default value to the table, a width of -1 means no setting

### DIFF
--- a/packages/arco-lib/src/components/Table/Table.tsx
+++ b/packages/arco-lib/src/components/Table/Table.tsx
@@ -82,6 +82,9 @@ export const exampleProperties: Static<typeof TablePropsSpec> = {
       displayValue: '',
       filter: false,
       componentSlotIndex: 0,
+      sorter: false,
+      ellipsis: false,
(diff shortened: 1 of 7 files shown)


packages/arco-lib/src/components/Table/Table.tsx
```diff
@@ -283,6 +296,9 @@ export const Table = implementRuntimeComponent({
     setColumns(
       cProps.columns!.map((column, i) => {
         const newColumn: ColumnProperty = { ...column };
+        if (newColumn.width === -1) {
+          Reflect.deleteProperty(newColumn, 'width');
```
tanbowensg, line 300: Since `newColumn.width` is checked below, do we have to delete it here? It is not good to mutate properties.

xzdry: changed

tanbowensg (approved):
LGTM

PR #588, opened 2022-09-01
feat(arco/Link): add onClick events to Link components

xzdry:


packages/arco-lib/src/components/Link.tsx
```diff
@@ -45,17 +46,30 @@ export const Link = implementRuntimeComponent({
...
 })(props => {
-  const { content, status, openInNewPage, ...cProps } = getComponentProps(props);
-  const { elementRef, customStyle, slotsElements } = props;
+  const { content, status, openInNewPage, href, ...cProps } = getComponentProps(props);
+  const { elementRef, customStyle, slotsElements, callbackMap } = props;
+
+  const handleClick = useCallback(
+    event => {
+      if (!href) {
+        event.preventDefault();
```
Yuyz0112, line 58: is this an always-want behavior?

xzdry: I'm not so sure
empty string is valid, its default behavior is to refresh the page
In some cases, the user wants a link-style button, but changing the style of the button can be cumbersome and not very reusable
So a click event is added to the link and the default reload behavior when the link href is empty needs to be prevented as I think this scenario is rarely used

xzdry: Or add a link style button
Temporarily close this pr

PR #589, opened 2022-09-02
fix(BooleanField): convert only the values that exist

xzdry:
Previously, booleanfield would accidentally convert undefined to true, now it only converts non-empty values

### DIFF
--- a/packages/editor-sdk/src/components/Widgets/BooleanField.tsx
+++ b/packages/editor-sdk/src/components/Widgets/BooleanField.tsx
@@ -16,7 +16,7 @@ export const BooleanField: React.FC<WidgetProps<BooleanFieldType>> = props => {
 
   useEffect(() => {
     // Convert value to boolean after switch from expression widget mode.
-    if (typeof value !== 'boolean') {
+    if (value && typeof value !== 'boolean') {
       onChange(true);
     }
   }, [onChange, value]);


packages/editor-sdk/src/components/Widgets/BooleanField.tsx
```diff
@@ -12,14 +12,15 @@ declare module '../../types/widget' {
...
 export const BooleanField: React.FC<WidgetProps<BooleanFieldType>> = props => {
-  const { value, onChange } = props;
+  const { value, onChange, path } = props;
 
   useEffect(() => {
     // Convert value to boolean after switch from expression widget mode.
-    if (typeof value !== 'boolean') {
-      onChange(true);
+    if (value && typeof value !== 'boolean') {
+      const convertValue = path[0] === 'ifCondition';
```
tanbowensg, line 20: I think we don't need treat `ifCondition` specially.

tanbowensg (changes requested):

packages/editor-sdk/src/components/Widgets/BooleanField.tsx
xzdry: Yes, here it should only be necessary to avoid accidental conversions     
changed

tanbowensg (approved):

PR #591, opened 2022-09-04
fix(runtime): fix #590, use an init set to check first time render

### LINKED ISSUE #590: Rethinking lifecycle management with slot trait's ifCondition property
**Version**

0.8 alpha

**Background**

In the latest version, we designed the new slot trait with more powerful features like slot props and the ifCondition property.

But when a component has loop slots and an application puts several components into these slots, how the ifCondition works may surprise you.

For example:

```
component slots: slot_content[key=1], slot_content[key=2], slot_content[key=3] // <- you can think [key=n] is a slot props

application:
  put component_1 into slot_content, ifCondition is `$slot.key === 1`
  put component_2 into slot_content, ifCondition is `$slot.key === 2`
  put component_3 into slot_content, ifCondition is `$slot.key === 3`
```

Actually, there will be 3*3=9 renders for each slot element, and the result is

```
component_1 + slot_content[key=1] = render
component_1 + slot_content[key=2] = skip
component_1 + slot_content[key=3] = skip
component_2 + slot_content[key=1] = skip
component_2 + slot_content[key=2] = render
component_2 + slot_content[key=3] = skip
component_3 + slot_content[key=1] = skip
component_3 + slot_content[key=2] = skip
component_3 + slot_content[key=3] = render
```

**What is expected?**

It may be reasonable to optimize the render times of this scenario, but in the meantime, we should expect every slot element works fine and the performance does not suffer.

**What is actually happening?**

In the `UnmountImplWrapper`, we will teardown the component's state from the store when it's unmounted. Currently, we are using the unmount feature to implement slot trait's ifCondition feature too, this makes it extremely hard to design when to teardown the store.

Yuyz0112:
This patch use a more solid way to check init state setup, instead of render count.

### DIFF
--- a/packages/runtime/src/utils/initStateAndMethod.ts
+++ b/packages/runtime/src/utils/initStateAndMethod.ts
@@ -15,7 +15,10 @@ export function initStateAndMethod(
   stateManager: StateManagerInterface,
   components: RuntimeComponentSchema[]
 ) {
-  components.forEach(c => initSingleComponentState(registry, stateManager, c));
+  components.forEach(c => {
+    stateManager.initSet.add(c.id);
+    initSingleComponentState(registry, stateManager, c);
+  });
 }
 
 export function initSingleComponentState(
@@ -26,6 +29,7 @@ export function initSingleComponentState(
   if (stateManager.store[c.id]) {
     return false;
   }
+
   let state = {};
   c.traits.forEach(t => {
     const tSpec = registry.getTrait(t.parsedType.version, t.parsedType.name).spec;
--- a/packages/runtime/src/utils/initStateAndMethod.ts
+++ b/packages/runtime/src/utils/initStateAndMethod.ts
@@ -16,8 +16,10 @@ export function initStateAndMethod(
   components: RuntimeComponentSchema[]
 ) {
   components.forEach(c => {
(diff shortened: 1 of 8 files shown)


packages/runtime/src/utils/initStateAndMethod.ts
```diff
@@ -15,7 +15,10 @@ export function initStateAndMethod(
   stateManager: StateManagerInterface,
   components: RuntimeComponentSchema[]
 ) {
-  components.forEach(c => initSingleComponentState(registry, stateManager, c));
+  components.forEach(c => {
+    stateManager.initSet.add(c.id);
```
tanbowensg, line 19: Maybe it is better to move this line into `initSingleComponentState`?

Yuyz0112: `initSingleComponentState` has been called in `UnmountImplWrapper` and `ImplWrapper`, while `initStateAndMethod` only called by the App and ModuleRenderer.

I'm not sure that scenarios need mark init set to true.

Yuyz0112:
BTW, I'm still testing this in dovetail.

Yuyz0112:
unit tests updated and is ready for merge

PR #593, opened 2022-09-05
 add slot key for table and remove tabs placeholder

xzdry:
1. add slot key for table
2. remove tabs placeholder

### DIFF
--- a/packages/arco-lib/src/components/Table/Table.tsx
+++ b/packages/arco-lib/src/components/Table/Table.tsx
@@ -436,8 +436,6 @@ export const Table = implementRuntimeComponent({
                 id: `${component.id}_${childSchema.id}_${index}`,
               };
 
-              console.log('id', `${component.id}_${childSchema.id}_${index}`);
-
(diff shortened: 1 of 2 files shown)


packages/arco-lib/src/components/Table/Table.tsx
```diff
@@ -172,7 +172,7 @@ export const Table = implementRuntimeComponent({
     state: TableStateSpec,
     methods: {},
     slots: {
-      content: {
```
tanbowensg, line 175: Although `td` is more proper than `content`, it is a breaking change. We'd better avoid breaking changes, before we have upgrade tool.

tanbowensg (changes requested):

packages/arco-lib/src/components/Table/Table.tsx
xzdry: ok,I have changed back to content

tanbowensg (approved):

PR #596, opened 2022-09-07
fix(editor): component will disappear after being dragged

tanbowensg:
Component will disappear when dragged to a component that does not have `content` slot.
Now the component will be add to the first slot of parent.


xzdry:
LGTM

xzdry (approved):
LGTM

PR #603, opened 2022-09-14
fix(editor): fix the bug that CodeEditor can not save code

tanbowensg:
Code Mode and CssEditor are broken because of this bug. I fixed it.

### DIFF
--- a/packages/editor/src/components/ComponentForm/StyleTraitForm/StyleTraitForm.tsx
+++ b/packages/editor/src/components/ComponentForm/StyleTraitForm/StyleTraitForm.tsx
@@ -267,6 +267,7 @@ export const StyleTraitForm: React.FC<Props> = props => {
                   mode="css"
                   defaultCode={style}
                   onBlur={v => changeStyleContent(i, v)}
+                  needRerenderAfterMount
                 />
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/ComponentForm/StyleTraitForm/StyleTraitForm.tsx
```diff
@@ -267,6 +267,7 @@ export const StyleTraitForm: React.FC<Props> = props => {
                   mode="css"
                   defaultCode={style}
                   onBlur={v => changeStyleContent(i, v)}
+                  needRerenderAfterMount
```
xzdry, line 270: `needRerenderAfterMount` always seems to be undefined. is this expected?

tanbowensg: Actually, it is `true` if no value is set.

xzdry: Sorry, I got confused

Yuyz0112 (approved):

PR #604, opened 2022-09-15
Fix the bug that api expression automatically changes to true

xzdry:
# How to reproduce this bug.
1. Create a api DataSource
2. Click the expression button to change the `disabled` property to an expression and enter something
3. onblur, `disabled` content will automatically change to `{{true}}` ,And an additional `"key": "value"` field is added to the json

# Reason
The reason for this bug is that api as a special DataSource will use `ApiForm` alone for rendering, but will also use `DataForm` for rendering as a common DataSource   
     
So when the disabled of the spec is modified to an `expression (e.g. {{xxx}})`, this spec is also passed to the `DataForm`.
But at this time the disabled in DataForm is not manually set to expression, so it will be rendered as `BooleanField` instead of `ExpressionWidget`   
  
And in BooleanFiled, it will turn the non-Boolean content to `true`, which will trigger the spec update again. This eventually causes the `ApiForm` to re-render and the disabled content to change to `{{true}}`. And the extra field `"key": "value"` will be added (the default behavior of DataForm)       

In this issue, I found that multiple api's DataForm will share the same state because the key of ObjectFiled component is the same, I also made a fix

# Solution
DataForm is not rendered when the type of DataSource is api

### DIFF
--- a/packages/editor-sdk/src/components/Widgets/ObjectField.tsx
+++ b/packages/editor-sdk/src/components/Widgets/ObjectField.tsx
@@ -38,7 +38,7 @@ export const ObjectField: React.FC<WidgetProps<ObjectFieldType>> = props => {
         return shouldRender(subSpec.conditions || [], value) ? (
           <SpecWidget
             component={component}
-            key={name}
+            key={name + component.id}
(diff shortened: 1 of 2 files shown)


tanbowensg (approved):
LGTM

PR #611, opened 2022-09-21
fix(docs/en): fix and format markdown tables

seruman:
Markdown tables were not rendered correctly due to spaces in header separators, removed spaces and formatted them.


tanbowensg:
Good job! Looks better!

tanbowensg (approved):

PR #613, opened 2022-09-22
refactor(runtime): export generateCallback and rename to runEventHandler

tanbowensg:
Some 3rd-party custom trait will need to call `generateCallback`, so I export it and rename it to a more proper name.


xzdry (commented):
LGTM

xzdry (approved):

PR #615, opened 2022-09-22
fix(editor): page freezes when fetch response is too large

tanbowensg:
Only render response when Accordion is open, so that the page will not freeze every time.

### DIFF
--- a/packages/editor/src/components/DataSource/ApiForm/Response.tsx
+++ b/packages/editor/src/components/DataSource/ApiForm/Response.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 import {
   HStack,
   Flex,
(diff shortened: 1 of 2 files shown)


packages/editor/src/components/DataSource/ApiForm/Response.tsx
```diff
@@ -57,19 +63,17 @@ export const Response: React.FC<Props> = props => {
...
               <CodeEditor
                 className={css`
                   width: 100%;
                   height: 100%;
                 `}
-                mode={{
-                  name: 'javascript',
-                  json: true,
-                }}
+                mode="text/plain"
```
Yuyz0112, line 74: maybe JSON is the expected display mode?

tanbowensg: ok, I changed back.

Yuyz0112 (approved):

PR #624, opened 2022-10-14
fix(editor): loss of state in ComponentList caused by tab switching

tuchg:
Before fix:
* The filter component information entered by the user is reset after leaving the tab, and the  experience is poor


tanbowensg:
LGTM

tanbowensg (approved):

PR #625, opened 2022-10-18
Arco/fix

xzdry:
reset to the first page if there is less data to show       

 synchronize clickedRow state when data changes

### DIFF
--- a/packages/arco-lib/src/components/Table/Table.tsx
+++ b/packages/arco-lib/src/components/Table/Table.tsx
@@ -215,6 +215,7 @@ export const Table = implementRuntimeComponent({
 
   const rowSelectionType = rowSelectionTypeMap[cProps.rowSelectionType];
   const currentChecked = useRef<(string | number)[]>([]);
+  const currentClickedRow = useRef<(string | number)[] | undefined>(undefined);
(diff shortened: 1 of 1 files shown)


packages/arco-lib/src/components/Table/Table.tsx
```diff
@@ -275,12 +276,31 @@ export const Table = implementRuntimeComponent({
...
       selectedRowKeys: selectedRows.map(r => r[rowKey]),
       selectedRows,
+      clickedRow,
     });
   }, [currentPageData, mergeState, rowKey]);
 
+  // If there is less data to display than the current page, reset to the first page
+  useEffect(() => {
+    if (useCustomPagination) return;
+    if (currentPageData.length <= Number(pageSize) * (currentPage - 1)) {
```
tanbowensg, line 292: Shouldn't it be `data.length < Number(pageSize) * (currentPage - 1)?`

xzdry: When the server-side paging option is not enabled, currentPageData equals data

tanbowensg (approved):
LGTM

PR #626, opened 2022-10-18
Feat/dry

xzdry:
1. add HTML component to show html content    
2. Add content and label slot to Timeline component

### DIFF
--- a/packages/arco-lib/src/components/Timeline.tsx
+++ b/packages/arco-lib/src/components/Timeline.tsx
@@ -64,26 +64,44 @@ export const Timeline = implementRuntimeComponent({
     properties: TimelinePropsSpec,
     state: TimelineStateSpec,
     methods: {},
-    slots: {},
+    slots: {
+      content: {
+        slotProps: Type.Object({
+          content: Type.String(),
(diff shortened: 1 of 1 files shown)


packages/runtime/src/components/core/HTML.tsx
```diff
@@ -0,0 +1,40 @@
...
+    slots: {},
+    styleSlots: ['content'],
+    events: [],
+  },
+})(({ elementRef, customStyle, html }) => {
+  return (
+    <div
+      ref={elementRef}
+      className={cx(customStyle?.content)}
+      dangerouslySetInnerHTML={{ __html: html }}
```
Yuyz0112, line 37: worth a try?
https://github.com/aknuds1/html-to-react

tanbowensg: Perhaps we'd better move this component to the project, not in runtime.

xzdry: The html components have been removed and will be placed in a custom component library

tanbowensg (approved):
LGTM

PR #629, opened 2022-10-19
Arco/fix

xzdry:
1. feat(arco/Select): add `addOption` and `setValue` methods to Select component
2. feat(arco/Tree): synchronize state when data changes
3. fix(arco/Tree): fix the error caused by changing data in multi-select mode
4. feat(arco): changing the defaultValue will trigger the onChange method(Temporarily added this behavior for `Input`、 `NumberInput`、`TextArea` and `Select`)


tanbowensg (approved):
LGTM

PR #630, opened 2022-10-26
feat(DataSource): add copy api function

xzdry:
copy


tanbowensg:
We'd better use `PasteComponentLeafOperation` to duplicate datasource. Because datasource is also a component under the hood. I expect that we can use one component tree to render both components and datasources. And that requires component and datasource use the same logic to duplicate.

PR #632, opened 2022-11-01
fix(CodeEditor): editing styles will override cssProperties

xzdry:
The reason for this bug is that the `UnControlled` component of `react-codemirror2` only updates [onBlur](https://github.com/scniro/react-codemirror2/blob/master/src/index.tsx#L204) in [componentDidMount](https://github.com/scniro/react-codemirror2/blob/master/src/index.tsx#L756), so subsequent calls to onBlur always get the cssProperties from the component mount in the closure context, thus overriding subsequent changes to the cssProperties


tanbowensg:
It is a litter tricky. Could you add a comment in code?

xzdry:
> It is a little bit tricky. Could you add a comment in code?

Added

tanbowensg (approved):

PR #634, opened 2022-11-03
Arco/fix

xzdry:
1.  feat(arco/Select): add trigger props to control the properties associated with the drop-down menu
2. feat(arco/Tree): add attributes to control the expansion of the tree 
3. feat(arco/Tree): selectedNode supports custom properties
4. feat(arco/Link): add onClick event and `asLink` properties are used to control whether or not to render as an `<a>` and add icon slot


tanbowensg (approved):
LGTM

PR #636, opened 2022-11-09
Auto detect component's DOM element when component doesn't implement elementRef or onRef

tanbowensg:
### Changes
1.  Feat: Auto detect component's DOM element 
2. Fix: mask does not update after component state changes.
3. add debounce time for update `EditorMask`

This is a fallback when component developer did not implement `elementRef` or `onRef`. The editor will traverse React Fiber Tree to find this component and its DOM Element, then use it to render `EditorMask`.

Because it uses React Fiber's implementation, it it not very stable. Developers should not rely on it.

### DIFF
--- a/packages/editor/src/components/EditorMaskWrapper/EditorMaskManager.ts
+++ b/packages/editor/src/components/EditorMaskWrapper/EditorMaskManager.ts
@@ -1,5 +1,7 @@
+import { debounce } from 'lodash';
 import { action, autorun, makeObservable, observable } from 'mobx';
 import React from 'react';
+import { traverse as traverseFiber } from 'react-fiber-traverse';
 import { EditorServices } from '../../types';
 
 type MaskPosition = {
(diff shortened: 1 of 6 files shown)


packages/editor/src/components/EditorMaskWrapper/EditorMaskManager.ts
```diff
@@ -163,14 +180,63 @@ export class EditorMaskManager {
...
+      }
     });
+    return map;
+  }
 
-    this.elementIdMap = elementIdMap;
+  // Auto find the components that does not have ref by React Fiber
+  // This function is a fallback.
+  // And because it uses React Fiber, it is not stable.
+  private autoFindNoRefElement() {
```
Yuyz0112: we can name this as `unsafe_autoFindNoRefElement`, which let people be careful when playing with this method.

tanbowensg: done

Yuyz0112 (approved):

PR #641, opened 2022-11-17
Revert "perf(editor): improve ComponentForm performance"

tanbowensg:
We cannot add key to ComponentForm because it will cause bug when user switch between components. The old component's property could overwrite new component.

This reverts commit 46d587827eac6829de85c1539da0ce1930f22c55.


xzdry (approved):
LGTM

MrWindlike (approved):

PR #643, opened 2022-11-23
Lazy render properties field

tanbowensg:
Chakra Accordion component does not support lazy mode so that the unexpanded categroy properties will stil be renderer.
I manually add lazy mode for it which can siginificantly improve the performance of components that have a lot of properties.

Also I close the animation of Accordion.

Before
截屏2022-11-23 上午11 53 47

After
 
截屏2022-11-23 上午11 44 45

### DIFF
--- a/packages/editor-sdk/src/components/Widgets/CategoryWidget.tsx
+++ b/packages/editor-sdk/src/components/Widgets/CategoryWidget.tsx
@@ -76,12 +76,12 @@ export const CategoryWidget: React.FC<WidgetProps<CategoryWidgetType>> = props =
   }, [spec.properties]);
 
   const onExpandedIndexChange = (index: ExpandedIndex) => {
-    console.log('index', index);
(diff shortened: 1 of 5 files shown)


packages/editor-sdk/src/components/Widgets/CategoryWidget.tsx
```diff
@@ -74,11 +75,51 @@ export const CategoryWidget: React.FC<WidgetProps<CategoryWidgetType>> = props =
     );
   }, [spec.properties]);
 
+  const onExpandedIndexChange = (index: ExpandedIndex) => {
+    console.log('index', index);
```
Yuyz0112: no console

tanbowensg: done

Yuyz0112 (approved):

PR #657, opened 2022-12-23
feat(editor): remove preview modal and add preview page

tanbowensg:
Remove preview modal and add preview page instead.

https://user-images.githubusercontent.com/12260952/209282750-49e83f2e-56ac-4939-8c53-068a29ded539.mov


MrWindlike:
It's probably better not to use a preview.html for the preview page, because users might launch the editor in the local project without preview.html setup.

tanbowensg:
> It's probably better not to use a preview.html for the preview page, because users might launch the editor in the local project without preview.html setup.

That's true... I will see if there is better way to do this.

PR #659, opened 2022-12-23
fix: fix a bug that the expression editor closing abnormally

### LINKED ISSUE #660: The popover settings menu of the events widget does not open after removing any component
**Version**
"@sunmao-ui/editor": "^0.8.4-alpha.5",

**Steps to reproduce**
Create two components and the events widget menu will open normally, delete either component and it will not open
**What is expected?**
Can be opened normally
**What is actually happening?**
events

xzdry:
Previously the expressions editor in the popover widget would automatically blur, making it impossible to edit. This was due to the fact that popover blur hides the content when it is closed.
Add the class `PREVENT_POPOVER_WIDGET_CLOSE_CLASS` to the expression modal to organise it to be closed when editing
fix #660


tanbowensg:
How does `PREVENT_POPOVER_WIDGET_CLOSE_CLASS ` work? Is it a chakra-ui class?

xzdry:
> How does `PREVENT_POPOVER_WIDGET_CLOSE_CLASS ` work? Is it a chakra-ui class?

It is a field customised by the sunmao popover widget to determine if the current popover class name contains this field, and thus whether the popover is closed.
There is another bug  [660](https://github.com/smartxworks/sunmao-ui/issues/660) that I suspect is also related to this

tanbowensg (approved):
LGTM

tanbowensg:
Conflict

xzdry:
> Conflict

Resolved
I have also made a fix for #660 under this pr. The cause of this bug was that the container was not getting the dom node. I have added a useEffect to ensure that the dom nodes are correctly fetched
PTAL

PR #661, opened 2022-12-26
add versions to localStorage trait

xzdry:
Add a `versions` field to localStorage, stored in the meta. The behavior of localStorage is now
- If the key already exists and the version is less than the existing version, then the existing localStorage value is read and synchronised to the global state.
- If the key already exists and the version is greater than the existing version, then the current localStorage is stored/updated using the given initial value and key
- For a previous value such as `#localStorage0@value:4000` it will become `#localStorage0@value:{"value":4000, "meta":{"version":0}}`.

### DIFF
--- a/packages/editor/__tests__/validator/mock.ts
+++ b/packages/editor/__tests__/validator/mock.ts
@@ -365,6 +365,7 @@ export const DynamicStateTraitSchema: ComponentSchema[] = [
         properties: {
           key: 'value',
           initialValue: {},
+          version: 0,
         },
       },
     ],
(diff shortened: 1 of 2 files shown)


tanbowensg:
Why `versions` is both in property and state? Could you provide an example about how to use `versions`?

xzdry:
> Why `versions` is both in property and state? Could you provide an example about how to use `versions`?

When the given version is greater than the current version in localStorage,the value in localStorage will be overwritten by the given initial value.
I replaced all versions with version,PTAL

tanbowensg (approved):

PR #662, opened 2022-12-28
 add view state button in component action menu

tanbowensg:
Checking component‘s state becomes easier.

https://user-images.githubusercontent.com/12260952/209752456-d0fbde3f-8d97-4368-8530-2958c3962f76.mov

### DIFF
--- a/packages/editor/src/components/StructureTree/ComponentNode.tsx
+++ b/packages/editor/src/components/StructureTree/ComponentNode.tsx
@@ -19,6 +19,7 @@ import { AppModel } from '../../AppModel/AppModel';
 import { ComponentId } from '../../AppModel/IAppModel';
 import { RootId } from '../../constants';
 import { RelationshipModal } from '../RelationshipModal';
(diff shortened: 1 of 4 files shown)


MrWindlike (approved):

packages/editor/src/components/StructureTree/ComponentNode.tsx
```diff
@@ -92,6 +92,14 @@ const ComponentNodeImpl = (props: Props) => {
     e.stopPropagation();
     setIsShowRelationshipModal(true);
   }, []);
+  const onClickShowState = useCallback(
+    (e: React.MouseEvent) => {
+      e.stopPropagation();
+      editorStore.setExplorerMenuTab(3);
```
xzdry: It might be better to use `ExplorerMenuTabs.State`

tanbowensg: ok

PR #667, opened 2023-01-09
fix(editor): allow user removing data source trait

tanbowensg:


MrWindlike:
If I remove the datasource trait for a Dummy component, does the schema leave an empty dummy component that cannot be removed on the editor?

tanbowensg:
> If I remove the datasource trait for a Dummy component, does the schema leave an empty dummy component that cannot be removed on the editor?

Yes. It will leave a empty dummy, but you can remove it.

MrWindlike:
When I remove the trait from dummy component, it shows the undefined category in the editor. It might be better to show it as an unknown category.

image

tanbowensg:
OK
截屏2023-01-18 上午11 05 23

MrWindlike (approved):

PR #668, opened 2023-01-10
feat: support default value for property spec

MrWindlike:
## Default Value
The `generateDefaultValueFromSpec` would use the `spec.default` to generate the properties default values.

## Init The Trait Properties
The properties values of the trait is now initialized preferentially using `exampleProperties`.

## The Component And Trait Form
The components and traits form no longer use `generateDefaultValueFromSpec` to generate properties values, because it can cause the editor shows value and actual value is different. For example, an enumeration type property whose value is `undefined` will appear as if the first option has been selected if the default value is generated in the form, which will mislead the user.


tanbowensg (commented):
LGTM. Could you add some `default` value for core components and traits?

MrWindlike:
OK, I have added some default values for the core traits properties.

tanbowensg (approved):
LGTM

PR #670, opened 2023-01-11
refactoring module forms

xzdry:
module form

The update contains the following

1.  module and app edit forms changed to modal
2. no longer use `spec.properties`, no longer support manual editing of module spec, remove edit spec button
3. `exampleProperties` are now displayed using `recordWidget`
4. add module event edit form
5. module properties type in Module widget is inferred from exampleProperties to generate specs

### DIFF
--- a/packages/core/src/module.ts
+++ b/packages/core/src/module.ts
@@ -15,9 +15,7 @@ export type Module = {
 
 type ModuleSpec = {
   properties: JSONSchema7;
-  events: {
-    name: string;
-  }[];
+  events: string[];
   stateMap: Record<string, string>;
 };
 
@@ -45,12 +43,6 @@ export function createModule(options: CreateModuleOptions): RuntimeModule {
       exampleProperties: options.metadata.exampleProperties || {},
     },
     spec: {
-      // `json-schema-editor` has a readonly root object by default,
-      // it provides two schema formats,array({type:'array'}) and object({type:'object'}).
-      // In sunmao, we only use the object schema, so we need to specify a default value here
-      // and silently fail when root selects array.
-      // This is a bit obscure, so should remove the array type of root from the json-schema-editor later
-      // TODO remove the array type of root from the json-schema-editor
       properties: { type: 'object' },
       events: [],
       stateMap: {},
--- a/packages/core/src/module.ts
+++ b/packages/core/src/module.ts
@@ -15,7 +15,9 @@ export type Module = {
 
 type ModuleSpec = {
   properties: JSONSchema7;
-  events: string[];
+  events: {
+    name: string;
+  }[];
   stateMap: Record<string, string>;
 };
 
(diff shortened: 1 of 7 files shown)


packages/core/src/module.ts
```diff
@@ -15,7 +15,9 @@ export type Module = {
 
 type ModuleSpec = {
   properties: JSONSchema7;
-  events: string[];
+  events: {
```
tanbowensg: Here we'd better use `string[]` as `events` type, just like components.

packages/editor-sdk/src/components/Widgets/ModuleWidget.tsx
```diff
@@ -19,6 +24,35 @@ declare module '../../types/widget' {
   }
 }
 
+const genSpec = (type: Types, target: any): TSchema => {
```
tanbowensg, line 26: Why not use a library to convert json to JsonSchema?

packages/editor-sdk/src/components/Widgets/ModuleWidget.tsx
```diff
@@ -65,6 +99,21 @@ export const ModuleWidget: React.FC<WidgetProps<ModuleWidgetType>> = props => {
     });
   };
 
+  const modulePropertiesSpec = useMemo<JSONSchema7>(() => {
+    const obj = mapValues(module?.metadata.exampleProperties, p => {
+      let type = Types.STRING;
+      let result;
+      if (isExpression(p)) {
```
tanbowensg: No need to check here. `deepEval` will check whether it is expression.

tanbowensg (changes requested):
Looks very nice, while there is a little need change.

packages/editor-sdk/src/components/Widgets/ModuleWidget.tsx
xzdry: I think this conversion function is easy to implement and easier to maintain  by implementing it yourself.
What do you think?

packages/core/src/module.ts
xzdry: Here I am concerned about the need to store more content in the future, such as descriptions or otherwise. The `string[]` type will be a breaking change when it is extended in the future

packages/editor-sdk/src/components/Widgets/ModuleWidget.tsx
xzdry: changed

packages/core/src/module.ts
tanbowensg: But this change itself is also a breaking change. And if events need more content, components spec also need change. Since component events spec is already `string[]` It will cause breaking change anyway.
Another reason is that If module events spec and component events spec are differenct, then the event code can not be reused between them.

packages/editor-sdk/src/components/Widgets/ModuleWidget.tsx
tanbowensg: ok.

packages/core/src/module.ts
xzdry: done

tanbowensg (approved):

PR #672, opened 2023-01-12
refactor(arco/radio): the default value type for radio should be any

xzdry:


tanbowensg:
Any is too loose for the value of radio. Any includes array and object, which can not be value of radio. `Type.Literal` is more proper.

xzdry:
> Any is too loose for the value of radio. Any includes array and object, which can not be value of radio. `Type.Literal` is more proper.

done,changed to `string|boolean|number`

tanbowensg (approved):

Issue #581, opened 2022-08-30
Move component and its children to module from application


tanbowensg:
**What problem does this feature solve?**

**What does the proposed API look like?**

PR #675, opened 2023-01-13
feat(editor): add extract module feature

tanbowensg:
# Feature
This feature can help user to extract sunmao components into module to refactor old code.
To do so, user need to choose a component as the root of module. This component and its children will be moved into new module. Sunmao will auto analyse the relationship of dependency between in-module components and out-module components. According to the relationship, sunmao will auto generate new module config and modify the properties of related components. Sunmao did most job of extracting module, user only need click to use.

related issue: #581 

https://user-images.githubusercontent.com/12260952/212246256-761af482-791e-4a7d-8453-f0faf4a3feab.mov

### DIFF
--- a/packages/runtime/src/services/StateManager.ts
+++ b/packages/runtime/src/services/StateManager.ts
@@ -18,7 +18,7 @@ dayjs.locale('zh-cn');
 type EvalOptions = {
   scopeObject?: Record<string, any>;
   overrideScope?: boolean;
-  fallbackWhenError?: (exp: string) => any;
+  fallbackWhenError?: (exp: string, err: Error) => any;
   // when ignoreEvalError is true, the eval process will continue after error happens in nests expression.
   ignoreEvalError?: boolean;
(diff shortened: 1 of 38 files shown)


packages/runtime/src/services/StateManager.ts
```diff
@@ -169,18 +171,10 @@ export class StateManager {
...
-        get(_, prop) {
-          return options.slotKey ? store[options.slotKey][prop] : undefined;
-        },
-      }
-    );
 
     options.scopeObject = {
       ...options.scopeObject,
-      $slot: redirector,
+      $slot: options.slotKey ? store[options.slotKey] : undefined,
```
tanbowensg, line 177: The original way pass a `Proxy` to `$slot`. But `$slot` doesn't work in module if using Proxy. So I pass `store[options.slotKey]` as `$slot` to fix this bug.

xzdry (approved):

PR #685, opened 2023-02-01
feat(Editor): add events history

xzdry:
Listening for all events and displaying logs

events


packages/runtime/src/services/apiService.ts
```diff
@@ -11,6 +11,7 @@ export function initApiService() {
       componentId: string;
       name: string;
       parameters?: any;
+      triggerId?: string;
```
tanbowensg: `triggerId` should be required because every event handler must have a trigger.

packages/editor/src/components/WarningArea/type.ts
```diff
@@ -0,0 +1,19 @@
...
+  name: string;
+  parameters: any;
+  triggerId?: string;
+};
+export type EventLog = {
+  time: string;
+  type: string;
+  target: string;
+  methodName: string;
+  triggered: string;
```
tanbowensg: Maybe `triggered` should also be called `triggerId`?

packages/editor/src/components/WarningArea/type.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export type Props = {
+  services: EditorServices;
+};
+
+export type Event = {
+  componentId: string;
+  name: string;
+  parameters: any;
+  triggerId?: string;
```
tanbowensg: `triggerId` should be required.

tanbowensg (changes requested):
I think event parameters should be rendered too. Considering parameters are too long to display in table cell, we can display part of parameter and use tooltip to show full parameters.

xzdry:
> I think event parameters should be rendered too. Considering parameters are too long to display in table cell, we can display part of parameter and use tooltip to show full parameters.

Completed
image

packages/editor/src/components/WarningArea/type.ts
xzdry: done

packages/editor/src/components/WarningArea/type.ts
xzdry: done

tanbowensg:
CI failed

xzdry:
> CI failed

fixed

tanbowensg (approved):

PR #694, opened 2023-03-20
Add logging of runtime related debugging information

xzdry:
1.  Record events and state changes triggered at runtime and return them to the user    
2. Add some common functions for logging debugging, such as printing to the console and saving to localStorage
image


packages/runtime/src/services/apiService.ts
```diff
@@ -16,6 +17,12 @@ export function initApiService() {
       fromId: string;
       eventType: string;
     };
+    /**
+     * @description: record debug info
+     */
+    debug: {
```
tanbowensg: Don't add a `debug` event and `DebugLoggerType`. Just add a `mergeState` event. Debugger can listen `uiMethod` directly.

xzdry: The following changes have been made：
1. Delete debug event and use mergeState event record the debug info for the merge state process
2. support module event 
3. Delete the debugger package and put it under runtime

tanbowensg (approved):

PR #696, opened 2023-03-22
fix(#679): ensure that ids are unique when copying components

### LINKED ISSUE #679: Copying the same element after refreshing the page will result in the same id
**Version**
@sunmao-ui/runtime: "^0.8.0",
@sunmao-ui/editor: "^0.8.0",

**Steps to reproduce**
1. Create an element with slots(e.g. `Stack`)
2. Insert an element and copy it, behaves normally
3. Refreshing the page and copying it again will result in two elements with the same id, which will cause the page to crash
copy_bug

**What is expected?**
When copied again, the same id should not appear

xzdry:
resolve #679

### DIFF
--- a/packages/editor/src/operations/leaf/component/pasteComponentLeafOperation.ts
+++ b/packages/editor/src/operations/leaf/component/pasteComponentLeafOperation.ts
@@ -9,16 +9,19 @@ export type PasteComponentLeafOperationContext = {
   component: IComponentModel;
 };
 
-let copyTimes = 0;
-
 export class PasteComponentLeafOperation extends BaseLeafOperation<PasteComponentLeafOperationContext> {
   private componentCopy!: IComponentModel;
 
   do(prev: AppModel): AppModel {
+    let copyTimes = 0;
+    const ids = prev.allComponents.map(c => c.id);
     this.context.component.allComponents.forEach(c => {
-      c.changeId(`${c.id}_copy${copyTimes}` as ComponentId);
+      let copyId = `${c.id}_copy${copyTimes}`;
+      while (ids.includes(copyId as ComponentId)) {
+        copyId = `${c.id}_copy${++copyTimes}`;
+      }
+      c.changeId(copyId as ComponentId);
     });
-    copyTimes++;
 
     this.componentCopy = this.context.component;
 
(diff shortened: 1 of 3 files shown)


packages/editor/src/operations/leaf/component/pasteComponentLeafOperation.ts
```diff
@@ -9,16 +9,19 @@ export type PasteComponentLeafOperationContext = {
...
   private componentCopy!: IComponentModel;
 
   do(prev: AppModel): AppModel {
+    let copyTimes = 0;
+    const ids = prev.allComponents.map(c => c.id);
     this.context.component.allComponents.forEach(c => {
-      c.changeId(`${c.id}_copy${copyTimes}` as ComponentId);
+      let copyId = `${c.id}_copy${copyTimes}`;
+      while (ids.includes(copyId as ComponentId)) {
+        copyId = `${c.id}_copy${++copyTimes}`;
```
tanbowensg: Don't use `++copyTimes`. It's hard to read.

packages/editor/src/operations/leaf/component/pasteComponentLeafOperation.ts
```diff
@@ -9,16 +9,19 @@ export type PasteComponentLeafOperationContext = {
   component: IComponentModel;
 };
 
-let copyTimes = 0;
-
 export class PasteComponentLeafOperation extends BaseLeafOperation<PasteComponentLeafOperationContext> {
   private componentCopy!: IComponentModel;
 
   do(prev: AppModel): AppModel {
+    let copyTimes = 0;
```
tanbowensg, line 16: If you move `copyTimes` to local scope, copyTimes will start from 0 every time user pastes. `copyTimes` should be global.

xzdry: Shared `copyTime` between components can cause copyTime to grow too fast and looks strangely     
For example，copy stack5 after copying text several times, the copytime becomes 13
image

packages/editor/src/operations/leaf/component/pasteComponentLeafOperation.ts
xzdry: changed

xzdry:
Added test, PTAL

tanbowensg (approved):

PR #697, opened 2023-03-22
refactor(ComponentSearch): remove choc-ui and replace it with select

xzdry:
Replace the `choc-ui autocomplete` component with the select (based on rc-select) provided by editor-sdk and remove the choc-ui package

Problems with choc-ui autocomplete include
1. performance not good, when there are many options there is a performance impact, when the page is refreshed or when the options item is switched up or down, all options items are re-rendered
2. the default filter is not friendly enough (can be solved by adding a filter function, but performance is not good, see 1)
3. sometimes the dropdown box will not pop up (no idea why at the moment)

This is a comparison between using choc-ui autocomplete and select(with the addition of the filter function)：    
choc autocomplete
component-search-choc

select

component-search

Normal use comparison：
component-search-compare

### DIFF
--- a/packages/editor-sdk/src/index.ts
+++ b/packages/editor-sdk/src/index.ts
@@ -4,3 +4,4 @@ export * from './constants/';
 export * from './models';
 export * from './types';
 export * from './utils';
+export * from './components/Select';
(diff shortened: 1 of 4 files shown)


tanbowensg (approved):
Good. choc-ui really has many problems.

PR #699, opened 2023-03-24
Feat/logs

xzdry:
improve `WarningArea` and `Logs` component：

- add display of `mergeState`
- optimisize the display of parameters
- add `eventType` field，the event column will be displayed using the event type first（eg: `onClick`）, and will only be set to the default value（eg: `Trigger Event`） if it is not available
- add resizable function to WarningArea
- optimize component jumps to UI_TREE, STATE, DATASOURCE

logs


packages/runtime/src/utils/runEventHandler.ts
```diff
@@ -38,6 +39,7 @@ export const runEventHandler = (
       name: evaledHandler.method.name,
       parameters: evaledHandler.method.parameters,
       triggerId,
+      eventType: eventType || (handler as any)?.type,
```
tanbowensg: Don't use `any`. `handler` has its own type `EventHandlerSpec`.

packages/runtime/src/services/debugger/type.ts
```diff
@@ -6,5 +6,5 @@ export type DebugLog = {
 };
 
 export type DebuggerHandler = {
-  onDebug: (currentState: Record<string, any>, debugLog: DebugLog) => void;
+  onDebug: (debugLog: DebugLog, currentState: Record<string, any>) => void;
```
tanbowensg, line 9: Why change the params order? Does it matters?

packages/runtime/src/utils/runEventHandler.ts
xzdry: changed

packages/runtime/src/services/debugger/type.ts
xzdry: The debuglog functions provided in the previous implementation were not aligned with onDebug, which resulted in the need to manually reorder them when using them   
before reordering
```ts
import { initSunmaoUI,printDebugInfo } from '@sunmao-ui/runtime';
  const ui = initSunmaoUI({
    // ...
    debugHandler:{
      onDebug:(state,log)=>{
        printDebugInfo(log,state);
      }
    }
  });
```
 after reordering
```ts
import { initSunmaoUI,printDebugInfo } from '@sunmao-ui/runtime';
  const ui = initSunmaoUI({
    // ...
    debugHandler:{
      onDebug: printDebugInfo
    }
  });
```
I think the impact is minimal, as this version is not yet released

tanbowensg (approved):

PR #702, opened 2023-03-28
fix(AppModel): update componentMap after copying top-level components

xzdry:
Fix: when copying a top-level component, it was not possible to continue copying its children because the corresponding child component had not been updated to the componentMap, resulting in an error when continuing to copy its children because it was not found.


packages/editor/__tests__/operations/component.spec.ts
```diff
@@ -48,6 +49,35 @@ describe('Component operations', () => {
...
+    );
+
+    const pasteComponent = [
+      appModel.getComponentById('stack3' as ComponentId)!.toSchema(),
+      appModel.getComponentById('stack5' as ComponentId)!.toSchema(),
+      appModel.getComponentById('text6' as ComponentId)!.toSchema(),
+    ];
+    const operation = genOperation(registry, 'pasteComponent', {
+      parentId: RootId,
+      slot: '',
```
tanbowensg, line 65: `slot` is empty?

xzdry: Yes, because the copy is of the top-level component

tanbowensg (approved):

PR #704, opened 2023-03-29
Fix/arco table

xzdry:
1. fix the problem that the `selectedRows` state was not synchronized correctly when the data changes when `checkCrossPage` is enabled for server-side paging
2. fix an error in tablePrimaryKeyWidget when columns is expression

### DIFF
--- a/packages/arco-lib/src/components/Table/Table.tsx
+++ b/packages/arco-lib/src/components/Table/Table.tsx
@@ -215,7 +215,7 @@ export const Table = implementRuntimeComponent({
   } = pagination;
 
   const rowSelectionType = rowSelectionTypeMap[cProps.rowSelectionType];
-  const currentChecked = useRef<any[]>([]);
+  const currentChecked = useRef<Record<string, unknown>[]>([]);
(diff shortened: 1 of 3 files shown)


packages/arco-lib/src/components/Table/Table.tsx
```diff
@@ -215,7 +215,7 @@ export const Table = implementRuntimeComponent({
   } = pagination;
 
   const rowSelectionType = rowSelectionTypeMap[cProps.rowSelectionType];
-  const currentChecked = useRef<(string | number)[]>([]);
+  const currentChecked = useRef<any[]>([]);
```
tanbowensg: Why use any?

xzdry: Since only the `selectedRowKeys` needed to be recorded before, the type can be determined.     
Now it has been changed to record `selectedRows` and the type of `table row` cannot be determined.  
I have changed the type to `Record<string,unkonwn>[]` and will infer the type when use it

tanbowensg (approved):

PR #711, opened 2023-05-05
fix(collapse): fix infinite loop caused by defaultValue as an array

xzdry:
This commit fixes an issue in the Collapse component where using an array as the defaultValue prop could cause an infinite loop. The issue was caused by a shallow comparison of the defaultValue prop in the useStateValue hook.

To fix the issue, a new useDeepCompareEffect hook was added, which allows for deep comparison of the defaultValue prop. This hook uses the lodash isEqual function to compare arrays.

### DIFF
--- a/packages/arco-lib/src/hooks/useStateValue.ts
+++ b/packages/arco-lib/src/hooks/useStateValue.ts
@@ -1,6 +1,7 @@
 import { RuntimeFunctions } from '@sunmao-ui/runtime';
 import { useState, useEffect } from 'react';
 import { SlotSpec } from '@sunmao-ui/core';
+import useDeepCompareEffect from './useDeepCompareEffect';
 
 export const useStateValue = <
   T,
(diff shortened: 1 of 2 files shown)


packages/arco-lib/src/hooks/useStateValue.ts
```diff
@@ -22,7 +23,7 @@ export const useStateValue = <
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  useEffect(() => {
```
tanbowensg, line 25: Is there any other components using this hook?

xzdry: All components that have the `updateWhenDefaultValueChanges` property will use this hook

tanbowensg: Then it is dangerous to modify it. It could cause unexpected problems. A new hook should be created.

xzdry: > Then it is dangerous to modify it. It could cause unexpected problems. A new hook should be created.

ok，perhaps it would be more appropriate to handle it in the collapse component

PR #717, opened 2023-08-29
feat(editor-sdk): add tooltip for too long ArrayTableRow

tanbowensg:
To solve the problem of ComponentId not being displayed completely in EventHandlerForm, a Tooltip has been added.

截屏2023-08-29 下午2 30 21


xzdry:
I noticed the title attribute here before, is tooltip necessary?

tanbowensg:
Title attribute requires user hover id for a while to show the full id, which is not very convenient.

xzdry (approved):

PR #721, opened 2023-08-30
feat(runtime): add watcher data source

tanbowensg:
Add a `watcher` data source, to watch expression changes, and trigger events.
With this data source, developer don't have to add `onChange` event for every component's state. It is also very useful to complicated expression.

The usage is like:
截屏2023-08-30 下午4 44 22

**Note that the `exp` property requires a string in expression format without `{{}}`.** This is to prevent the `exp` being evaled when compiling template.

### DIFF
--- a/packages/runtime/src/components/core/Watcher.ts
+++ b/packages/runtime/src/components/core/Watcher.ts
@@ -1,0 +1,42 @@
+import { Type } from '@sinclair/typebox';
+import {
+  CORE_VERSION_V2,
+  CoreComponentName,
+  PRESET_PROPERTY_CATEGORY,
+} from '@sunmao-ui/shared';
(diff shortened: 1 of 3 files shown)


MrWindlike:
1. I'm not sure if we need to provide a Watcher at runtime. Because when listening for state changes to do something, it is easier to use it incorrectly.
2. Why not just pass in the expression and listen for changes in its value directly in Watcher, instead of compiling it in Watcher? Writing plain strings has no editor prompt and is error-prone
3. Can it support watch multiple states?

tanbowensg:
@MrWindlike I changed implementation. Now the component implementation get raw `exp` from `component.properties`, so that user can have editor prompt when entering `exp`.

MrWindlike (approved):

PR #725, opened 2023-09-04
feat(editor): support use tags to filter data source

tanbowensg:
### Feature
Support categorizing DataSources using Tags.

https://github.com/smartxworks/sunmao-ui/assets/12260952/ead0f725-937a-4c40-8899-07655a2b7775

### Storage
The data is stored in the metadata of the Application Schema, look like: 

截屏2023-09-04 下午6 11 19

To add a new Tag, you can directly modify the Schema and add a new key in the `componentsTagMap` with an `[]` as the value.

### DIFF
--- a/packages/editor/src/components/DataSource/DataSourceList.tsx
+++ b/packages/editor/src/components/DataSource/DataSourceList.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import {
   VStack,
   Spacer,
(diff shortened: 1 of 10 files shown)


packages/editor/src/components/DataSource/DataSourceList.tsx
```diff
@@ -30,8 +31,25 @@ export const DataSourceList: React.FC<Props> = props => {
...
+  const filteredDataSources = useMemo(() => {
+    if (!tagMap || checkedTags.length === 0) {
+      return dataSources;
+    }
+
+    const dataSourceIds = checkedTags.reduce<string[]>((result, curr) => {
+      return result.concat(tagMap[curr]);
+    }, []);
+
+    return uniq(dataSourceIds)
```
MrWindlike: Maybe change it to something like this to be more simple and clear:
```js
return dataSources.filter(ds=> dataSourceIds.includes(ds.id));
```

MrWindlike (approved):

